PR #1, opened 2023-05-01
Import hex code

tcharding:
Import the following files
    
- `buf_encoder.rs`: from `internals/src/hex/buf_encoder.rs`
- `display.rs`: from `internals/src/hex/display.rs`
- `mod.rs`: from `internals/src/hex/mod.rs`
- `parse.rs`: from `hashes/src/hex.rs`
    
The files above can be diffed to see the changes introduced, which should be minimal.

Includes all standard infrastructure: GitHub actions, rustfmt, clippy, githooks

Does not include lock files (minimal and recent). I had a go at this but we only have two dependencies (`core2` direct, and `memchr` transient) so it doesn't seem necessary.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -1,0 +1,104 @@
+on:
+  push:
+    branches:
+      - master
+      - 'test-ci/**'
+  pull_request:
+
+name: Continuous integration
+
+jobs:
+  Stable:
+    name: Test - stable toolchain
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v3
+      - name: Checkout Toolchain
+        # https://github.com/dtolnay/rust-toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Running test script
+        env:
+          DO_LINT: true
+          DO_DOCS: true
+          DO_FEATURE_MATRIX: true
+        run: ./contrib/test.sh
+
+  Beta:
+    name: Test - beta toolchain
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v3
+      - name: Checkout Toolchain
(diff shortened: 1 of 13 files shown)


tcharding:
@apoelstra can you enable github actions please. Can we have multiple maintainers on this crate? Happy to volunteer.

apoelstra:
Added you and @Kixunil as admins on the github repo. Let me know if either of you want to be crates.io maintainers as well.

apoelstra:
I don't know what I need to do to enable Actions -- I think we need to merge a PR with actions first.

apoelstra:
Done review. utACK 1ccd52b4f88effe5fc79b79bc56f82db905be6ae. Diffed against the most recent rev of the rust-bitcoin "hex crate" PR that I'd ACKed. Basically just formatting and import changes, and using `track_caller` unconditionally.

Will setup a local CI then ACK and merge this, to see if that causes github actions to engage. Then we can do a followup PR with responses to whatever @Kixunil's review turns up.

apoelstra:
I have one nit, which is that clippy on nightly outputs:

```
error: this `impl` can be derived
  --> src/lib.rs:66:1
   |
66 | / impl Default for Case {
67 | |     fn default() -> Self { Case::Lower }
68 | | }
   | |_^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derivable_impls
   = note: `-D clippy::derivable-impls` implied by `-D warnings`
   = help: remove the manual implementation...
help: ...and instead derive it...
   |
56 + #[derive(Default)]
57 | pub enum Case {
   |
help: ...and mark the default variant
   |
60 ~     #[default]
61 ~     Lower,
   |
```
I think there is some way to configure the clippy MSRV, but for now I'll skip clippy and we can address this in a followup.

apoelstra:
Hehe, so the good news is that Github Actions work. The bad news is that our CI apparently also fails if clippy fails, so we have a red X due to that `#[default]` thing. (Is this available in 1.48? I guess we should try..)

tcharding:
Strange, I included a clippy config file in this PR and I get no clippy errors locally on master with this merged - or did you fix it somehow?

PR #7, opened 2023-05-05
Rename `HexIterator` to `Iter`

tcharding:
A few observations:

- The `HexIterator` is an adaptor iterator, iterating over a hex string yielding bytes.
- The `HexIterator` is passed to a function called `from_bytes_iter`.
- It is unlikely that we will ever have an adaptor iterator iterating over a bytes an yielding two hex digits.

The name `Iter` seems like a good candidate, if re-exported at the crate root this allows usage as `hex::Iter::new("deadbeef")`.

Rename `HexIterator` to `Iter` and re-export it at the crate root.

Inspired by https://github.com/rust-bitcoin/rust-bitcoin/pull/1833#issuecomment-1535149929

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,7 +37,7 @@ pub mod display;
 pub mod parse;
 
 pub use display::DisplayHex;
-pub use parse::{Error, FromHex};
+pub use parse::{Error, FromHex, Iter};
 
 /// Reexports of extension traits.
 pub mod exts {
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -53,27 +53,27 @@ pub trait FromHex: Sized {
         I: Iterator<Item = Result<u8, Error>> + ExactSizeIterator + DoubleEndedIterator;
 
     /// Produces an object from a hex string.
-    fn from_hex(s: &str) -> Result<Self, Error> { Self::from_byte_iter(HexIterator::new(s)?) }
+    fn from_hex(s: &str) -> Result<Self, Error> { Self::from_byte_iter(Iter::new(s)?) }
 }
 
 /// Iterator over a hex-encoded string slice which decodes hex and yields bytes.
-pub struct HexIterator<'a> {
+pub struct Iter<'a> {
     /// The `Bytes` iterator whose next two bytes will be decoded to yield
     /// the next byte.
     iter: str::Bytes<'a>,
 }
 
-impl<'a> HexIterator<'a> {
-    /// Constructs a new `HexIterator` from a string slice.
+impl<'a> Iter<'a> {
+    /// Constructs a new `Iter` from a string slice.
     ///
     /// # Errors
     ///
     /// If the input string is of odd length.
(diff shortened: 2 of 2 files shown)


tcharding:
Oh and this is an example of my new effort to _really_ make sure I understand why reviewers suggest things instead of doing them blindly. If I ever misrepresent your suggestion please say so so I can correct my understanding. Thanks.

apoelstra:
Hmm, I appreciate you writing stuff out explicitly because it makes it clear what we're doing here.

Given that this is an adaptor iterator, maybe we ought to add a `Hexed` trait that we would implement on all existing byte iterators. Then users can add `.hexed()` to their iterator chains rather than typing `Iter::new`.

Maybe `Hexed` should be called `IteratorExt`. Unsure. @Kixunil what are your thoughts?

Meanwhile I think this specific PR, which just renames `HexIterator` to `Iter` and re-exports it at the root, is good to go.

apoelstra (approved):
ACK fb37f96dcb1f4f0a59663e39a6e9447321ecc2a9

Kixunil:
> * It is unlikely that we will ever have an adaptor iterator iterating over a bytes an yielding two hex digits.

I'm not sure about this one. I wanted both encoding and decoding iterator (as well as read/write) for a long time ago and would like to revisit this. Something like `DecodeIter` would be better.

If I saw `.hexed` in the code I wouldn't know if it's encoding or decoding. I'd prefer `encode_hex` and `decode_hex`.

apoelstra:
@Kixunil cool, concept ACK having distinct `encode_hex`/`decode_hex`. Or better, `bytes_to_hex` and `hex_to_bytes`?

Then actually the iterators could be called `HexToBytes` and `BytesToHex`, which would match the function names and still (IMHO) fairly clearly hint that they are iterator adaptors.

Kixunil:
Yeah, those soud better. It's a bit tempting to throw in `Iter` suffix but maybe not needed. (e.g. `str::Bytes` don't have it)

tcharding:
I slightly prefer having explicit `Iter`, "bytes" is ambiguous (slice, iterator, array), I often have to look it up. `HexToBytesIter` is brain dead simple to see what it is.

It doesn't need explaining now but I still don't see how we would provide such an iterator because a "bytes iterator" yields bytes, one at a time (at least in my mind) but conversion from hex needs two bytes. Anyways, that is why I wrote the statement that we wouldn't have such a thing. Happy to be proved wrong.

apoelstra:
Ok, I'm cool with the `Iter` suffix.

apoelstra:
> It doesn't need explaining now but I still don't see how we would provide such an iterator because a "bytes iterator" yields bytes, one at a time (at least in my mind) but conversion from hex needs two bytes

When going from bytes to hex we would store each byte for two `next` calls before requesting a new one from the underlying iter. When going the other direction, we'd consume two characters at once when yielding a single byte (and if we only consumed one we'd return an error).

tcharding:
Simple now I read it :)

PR #9, opened 2023-05-10
Add encoding iterator

tcharding:
This PR was inspired by discussion on https://github.com/rust-bitcoin/hex-conservative/pull/7

Add an encoding iterator `BytesToHexIter`. On the way there do a bunch of improvements, note carefully please the changes to error handling in commit: `51ba756 Improve error types`.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -2,6 +2,7 @@
 
 //! Iterator that converts hex to bytes.
 
+use core::iter::FusedIterator;
 use core::str;
 #[cfg(feature = "std")]
 use std::io;
@@ -65,7 +66,7 @@ impl<'a> ExactSizeIterator for HexToBytesIter<'a> {
     fn len(&self) -> usize { self.iter.len() / 2 }
 }
 
-impl<'a> core::iter::FusedIterator for HexToBytesIter<'a> {}
+impl<'a> FusedIterator for HexToBytesIter<'a> {}
 
 #[cfg(any(feature = "std", feature = "core2"))]
 impl<'a> io::Read for HexToBytesIter<'a> {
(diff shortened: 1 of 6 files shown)


apoelstra:
I'm a little sad to see the entire `internals` crate being pulled in for a single instance of `write_err`. I'd suggest copying that macro into this crate, or even inlining it in the single location that it's used.

Other than that, ACK 0a3a01731788a2ae0ca19d6938e2a5e819d8c98d

tcharding:
Changes in force push:
- Removed dependency on `internals`, copied the `write_err` macro instead.
- Rebased and put formatting changes in the correct patch.

tcharding:
This branch is now used in the demo usage branch: https://github.com/rust-bitcoin/rust-bitcoin/pull/1833

apoelstra (approved):
ACK 34b03d4c0419e1f989d7c9f6f57c7686960a4ec7

apoelstra:
cc @Kixunil will let you take a look before merging this

tcharding:
Hey @Kixunil do you have time to take a look at this one, would be cool to get it in a pre-release.

src/iter.rs
```diff
@@ -56,6 +48,18 @@ impl<'a> Iterator for HexToBytesIter<'a> {
     }
 }
 
+impl<'a> DoubleEndedIterator for HexToBytesIter<'a> {
+    fn next_back(&mut self) -> Option<Result<u8, HexToBytesError>> {
+        let lo = self.iter.next_back()?;
+        let hi = self.iter.next_back().unwrap();
+        Some(chars_to_hex(hi, lo))
+    }
+}
```
Kixunil, line 58: Reminds me we should also override `nth` and `nth_back`.

src/iter.rs
```diff
@@ -85,3 +86,158 @@ fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, HexToBytesError> {
...
+            }
+            None => self.iter.next().map(|b| {
+                let (high, low) = byte_to_hex_chars(b);
+                self.low = Some(low);
+                high
+            }),
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) { (self.len(), Some(self.len())) }
```
Kixunil, line 126: `self.len()` normally calls into `size_hint`, not the other way around. It's not wrong but might be a bit better to keep the standard.

src/iter.rs
```diff
@@ -85,3 +86,158 @@ fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, HexToBytesError> {
...
+where
+    I: ExactSizeIterator + FusedIterator + Iterator<Item = u8>,
+{
+    /// Constructs a new `BytesToHexIter` from a byte iterator.
+    pub fn new(iter: I) -> BytesToHexIter<I> { Self { iter, low: None } }
+}
+
+impl<I> Iterator for BytesToHexIter<I>
+where
+    I: ExactSizeIterator + FusedIterator + Iterator<Item = u8>,
```
Kixunil, line 108: This shouldn't require `ExactSizeIterator`.

Kixunil (commented):
Among these things, I think we could also change `HexToBytesIter` to contain a user-defined iterator instead of forcing bytes, probably as a separate PR.

src/parse.rs
```diff
@@ -99,20 +104,57 @@ impl<'a> core::iter::FusedIterator for HexToBytesIter<'a> {}
...
     {
         iter.collect()
     }
 }
 
+/// Hex decoding error.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum HexToBytesError {
+    /// Non-hexadecimal character.
+    InvalidChar(u8),
```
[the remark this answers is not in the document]
tcharding: I tried putting `#[non_exhaustive]` on enum error variants so we can do future changes like you mention but it seems to make the error type basically unusable because then one cannot match on the variants.

src/parse.rs
```diff
@@ -14,13 +14,18 @@ use crate::alloc::vec::Vec;
 
 /// Trait for objects that can be deserialized from hex strings.
 pub trait FromHex: Sized {
+    /// Error type returned while parsing hex string.
+    type Error: From<HexToBytesError> + Sized + fmt::Debug + fmt::Display;
```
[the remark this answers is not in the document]
apoelstra: We have these sorts of constraints in rust-miniscript. Without them you can't `.expect` on generic errors. (Though maybe this is okay for this library, because like `FromStr`, its operations are small and will typically only occur on concrete types.)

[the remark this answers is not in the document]
tcharding: Thanks, I removed the `Send`, `Debug`, and `Display`. Cannot work out how to use the `Into` suggestion?

src/parse.rs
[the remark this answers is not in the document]
apoelstra: You can definitely match on `#[non_exhaustive]` errors. You just need to stick a `_` clause on to deal with potential future errors.

Having said this, `#[non_exhaustive]` would be insufficient to *replace* one variant with another, or to change its internal fields.

@Kixunil when we return these errors from iterators we may not know the "position", and accumulating multiple erroneous chars would require allocation. I'm not sure it's worth it for an error variant like this, where we give the user a character they can easily search their string for (and where the error will almost always indicate "this string is not hex at all, exact error is irrelevant").

src/parse.rs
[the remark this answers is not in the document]
tcharding: I don't know why `FromStr` and `TryFrom` don't require `Debug`, an error type seems pretty useless if it does not implement even `Debug`?

src/iter.rs
tcharding: Thanks, fixed.

src/parse.rs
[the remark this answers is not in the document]
tcharding: > You can definitely match on #[non_exhaustive] errors. You just need to stick a _ clause on to deal with potential future errors.
>
> Having said this, #[non_exhaustive] would be insufficient to replace one variant with another, or to change its internal fields.

Oh I meant putting `#[non_exhaustive]` on each variant. This idea came from that error blog post you linked to a while back.

src/iter.rs
tcharding: I added an issue.

tcharding:
Changes in force push:
- put typo fix in separate patch
- changed associated error type trait bounds (I may have this wrong still, has `Debug` because I rekon its better and uses `From` because I could not work out the suggestion to use `Into`)
- reduces trait bounds on iterator generics on `HexToBytesIter`

tcharding:
Rebased to pick up recent fuzzing changes and added a `decode` fuzz test for good measure.

tcharding:
> I'm a little sad to see the entire `internals` crate being pulled in for a single instance of `write_err`.

I was playing around with creating `bitcoin-units` crate today and moved some stuff to `internals`. What's your heuristic on when its okay to pull in `internals` and when do we just inline/duplicate stuff?

src/parse.rs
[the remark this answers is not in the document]
apoelstra: Ah, yeah, I see. Agreed, I'd rather not do this. We'd have to implement a ton of accessor methods and stuff. Let's just get it right the first time :)

apoelstra:
Roughly
* If we need more than 1 or 2 functions, **or**
* we already depend on `bitcoin` or something that has `internals` in its dep tree

But I don't have a hard/fast rule.

src/parse.rs
```diff
@@ -150,76 +192,74 @@ impl_fromhex_array!(512);
...
-    /// Non-hexadecimal character.
-    InvalidChar(u8),
-    /// Purported hex string had odd length.
-    OddLengthString(usize),
-    /// Tried to parse fixed-length hash from a string with the wrong type (expected, got).
+pub enum HexToArrayError {
+    /// Conversion error while parsing hex string.
+    Conversion(HexToBytesError),
+    /// Tried to parse fixed-length hash from a string with the wrong length (got, want).
     InvalidLength(usize, usize),
```
apoelstra, line 121: In 118ee3acb1e3c431497f6c9a753458d38d2d1a7c

I have a mild preference for making `got` and `want` named fields. But this is fine.

apoelstra (approved):
ACK 47c0d0ab88a347d0182afc959abecfb1e8fe02f9

src/parse.rs
tcharding: I've done that somewhere in the last week, I remember because I used `expected` and `got`, and I wanted to go back and change it to `expected` and `actual` (I favour got and want in unit tests but thought the longer ones were better for field names, do you have a preference?)

tcharding:
Added got/want field names as a separate patch because the `InvalidLength` variant was not previously touched by this PR. (See git log of the last patch for serendipity :)

src/parse.rs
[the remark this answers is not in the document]
Kixunil: There are some niche use cases for not requiring `Debug`, anyway it doesn't prevent you from calling `expect` because the generic can just add `Foo::Error: fmt::Debug` bound.

src/parse.rs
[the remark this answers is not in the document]
Kixunil: We may not know the position but we can use `Option<usize>` for it. (Although I believe we always can know it.)

Yes, reporting multiple errors should be opt-in because of required allocation. I've started working on something that looks pretty neat to me. What I have in mind is rustc-like beautiful errors.

apoelstra (approved):
ACK e681d53e61c4804ebf900b6d042391c1f650a248

apoelstra:
@Kixunil could you take another look at this?

Also, what do we need to do to get a 0.1 release out?

tcharding:
I'll undraft #12 if/when this goes in and from my perspective there is nothing else to do, we can release. @kixunil?

tcharding:
I'll undraft #12 if/when this goes in and from my perspective there is nothing else to do, we can release. @kixunil?

tcharding:
I just attempted to upgrade `rust-esplora-client` while trying to upgrade `bdk` and had to use `bitcoin_internals` to convert a vector to hex. We totally rugged everyone by not releasing this crate before we did the v0.30.0 release. The alternative, to add a dependency on `hashes` just to get `hashes::hex::to_hex` to super ugly when the same file is using `use bitcoin::hashes::Foo`.

apoelstra:
@tcharding 100% agreed. In rust-elements I would up just copy/pasting the `hex` module from bitcoin-hashes into `elements` (and I delayed upgrading for like six months waiting for this to be released). In rust-miniscript I depended on bitcoin-private.

tcharding:
Rebased, no other changes. (Sorry for the noise, done to get the release PR upto date.)

tcharding:
I'm getting super tired of using CI to find my mistakes - why the f**k does `cargo clippy --all-targets --all-features` not work at the crate root when we have `fuzz` as part of the workspace?

tcharding:
Please baby Jesus, merge this PR already :)

tcharding:
Full disclosure, until 10 minutes ago this PR was unmergable because I had not actually removed all the ExactSizedIterator stuff that @Kixunil comment on above weeks ago - my fail.

src/iter.rs
```diff
@@ -92,3 +93,154 @@ fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, HexToBytesError> {
...
+                let (high, low) = byte_to_hex_chars(b);
+                self.low = Some(low);
+                high
+            }),
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (min, max) = self.iter.size_hint();
+        (min * 2, max.map(|max| max * 2))
```
apoelstra, line 135: In fdeb91bf87eb44ea09fe073f57151fd5399363f4:

I'm a little confused why you removed the `self.low` calculus from this.

src/iter.rs
```diff
@@ -92,3 +93,154 @@ fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, HexToBytesError> {
...
+                let (high, low) = byte_to_hex_chars(b);
+                self.low = Some(low);
+                high
+            }),
+        }
+    }
+}
+
+impl<I> FusedIterator for BytesToHexIter<I> where
+    I: ExactSizeIterator + FusedIterator + Iterator<Item = u8>
```
apoelstra, line 159: In fdeb91bf87eb44ea09fe073f57151fd5399363f4:

I don't think we need the `ExactSizeIterator` bound here.

Separately, we should impl `ExactSizeIterator` on this type (and **that** will need an `ExactSizeIter` bound).

tcharding: Yes and yes, done. Thanks.

src/iter.rs
tcharding: Oh yes, no idea why I did that. I've re-written this to use `self.low`.

tcharding:
Thanks for the review @apoelstra, was a bit sad to see I'd made so many mistakes after pushing for merge :(

apoelstra (approved):
ACK 33e81052babc5712adfe671b2df01a239e936aab

PR #10, opened 2023-05-11
Add a mixed case example/test

tcharding:
This library can parse hex strings of either case. Also we do not care if a hex string contains a mix of uppercase and lowercase, however we only output in a single case. This means mixed case strings do not roundtrip.

Add an example and a unit test that shows parsing of mixed case hex strings.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,13 @@
 //! println!("An array as lower hex: {:x}", a.as_hex());
 //! // And for vecs since `Vec` derefs to byte slice.
 //! println!("A vector as upper hex: {:X}", v.as_hex());
+//!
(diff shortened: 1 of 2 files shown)


apoelstra:
FWIW Bitcoin Core accepts mixed-case hex.

apoelstra (approved):
ACK f1d2ea097f27da5a8444fbdfe221810d0828171e

tcharding:
I think its fine to leave checking case to higher layers up the stack, if we added `from_hex_no_mixed_case` it would be ugly and no one would use it anyways.

tcharding:
Lets just merge this one, its been ack'd with no nack for two weeks.

PR #15, opened 2023-06-20
Add changelog and prep for release

tcharding:
Prepare for release by doing:

- Bump the version to 0.1.0
- Add changelog file and populate an entry for v0.1.0

Now includes initial patch to remove the todo (see suggested in discussion below). FTR I introduced the todo in #9.

## Merge before release

- #20 
- #21

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -207,7 +207,6 @@ mod tests {
     #[test]
     fn decode_iter_backward() {
         let hex = "deadbeef";
-        // TODO: Confirm that this is what we want, the digits are still ordered (high, low)?
         let v = vec![0xef, 0xbe, 0xad, 0xde];
 
         for (i, b) in HexToBytesIter::new(hex).unwrap().rev().enumerate() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+# 0.1.0 - 2023-06-20 Initial Release
+
+- [Import](https://github.com/rust-bitcoin/hex-conservative/pull/1) code from the `bitcoin_hashes` and `bitcoin-internals` crates.
+- [Add `Iterator` implementations](https://github.com/rust-bitcoin/hex-conservative/pull/9)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hex-conservative"
-version = "0.0.1"
+version = "0.1.0"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 repository = "https://github.com/rust-bitcoin/hex-conservative"
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -10,7 +10,7 @@ cargo-fuzz = true
 
 [dependencies]
 honggfuzz = { version = "0.5.55", default-features = false }
-hex = { path = "..", package = "hex-conservative", version = "0.0.1" }
+hex = { path = "..", package = "hex-conservative", version = "0.1.0" }
 
 [[bin]]
 name = "hex"
--- a/fuzz/generate-files.sh
+++ b/fuzz/generate-files.sh
@@ -21,7 +21,7 @@ cargo-fuzz = true
 
 [dependencies]
 honggfuzz = { version = "0.5.55", default-features = false }
-hex = { path = "..", package = "hex-conservative", version = "0.0.1" }
+hex = { path = "..", package = "hex-conservative", version = "0.1.0" }
 EOF
 
 for targetFile in $(listTargetFiles); do


tcharding:
@Kixunil we are going to go ahead and merge this and release even though you are on holiday. I don't believe there are any urgent improvements required and its only a v0.1.0 release anyways, we can release again if needed.

tcharding:
My sleep deprive brain cannot tell me what additionally needs doing to fix CI, will re-try tomorrow unless you know off the top of your head please @apoelstra? Something is looking for version 0.0.1 (this PR includes bump in the manifest to 0.1.0).

apoelstra:
Without digging into it, it looks to me like we're writing `0.0.1` somewhere and `0.1.0` somewhere else. These are different strings, which maybe you couldrn't tell when sleep-deprived?

tcharding:
I think technically that's dyslexia not sleep dep :)

Thanks though, I'll fix it up.

tcharding:
Turns out I'm not retarded and can grep, I was just being a bit silly - this needed rebasing to pick up the fuzz directory.

tcharding:
@Kixunil can you please take a look at this over the next few days. Lets release this crate already.

tcharding:
Full disclosure, until 10 minutes ago this PR was unmergable because I had not actually removed all the `ExactSizedIterator` stuff that @Kixunil comment on above weeks ago - my fail.

apoelstra:
Should this be moved out of draft?

src/iter.rs
```diff
@@ -92,3 +93,161 @@ fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, HexToBytesError> {
...
+
+        for (i, b) in HexToBytesIter::new(hex).unwrap().enumerate() {
+            assert_eq!(b.unwrap(), v[i]);
+        }
+    }
+
+    #[test]
+    fn decode_iter_backward() {
+        let hex = "deadbeef";
+        // TODO: Confirm that this is what we want, the digits are still ordered (high, low)?
```
apoelstra, line 210: In 33e81052babc5712adfe671b2df01a239e936aab:

Yes, this is what we want. Thank you for adding a test.

apoelstra (approved):
utACK aa03b64a741e3b7beca1af4f31125e23e296d282

apoelstra (approved):
ACK aa03b64a741e3b7beca1af4f31125e23e296d282

apoelstra:
@Kixunil I think I'm gonna merge and release this. I'll give you a few days to jump in, in case you want to take a look.

apoelstra:
Meanwhile I'll see if I can use it to replace the hex code in rust-elements, as a test.

tcharding:
Changes in force push:
- rebased on master
- removed the todo commented on above

Also re-wrote PR description

tcharding:
Note please, I just noticed that we have trait bounds on the `FromHex` associated error but IIRC @Kixunil mentioned that he did not like these. Not sure if they got back in through a mistake on my behalf or if I'm thinking of some other place/time/thing.

apoelstra:
I also noticed these, and do not like them, but without them it's not clear to me how we would implement this trait on reversible hashes.

In practice this covers all the slice iterators so it's not the end of the world.

apoelstra:
Ah, without reversibility we can just yield bytes in the normal order and then reverse them in-place. That's easy.

But without `ExactSizeIter` then we can't enforce that our iterators have an even number of chars, and have to decide what to do if there is an odd number.

tcharding:
I'll crack on with those suggestions so we can get them in before we release.

apoelstra:
I think my last cut-off "At the top level we should" suggestion was going to be something about providing an `Error` type that included some common "parsing failed" variants but I decided against it. Users can define their own types for their own errors.

tcharding:
> I think we should rename FromHex::Error to FromHex::Err so that it matches the error variant for FromStr.

Done, I've got a feeling @Kixunil isn't going to like it, flagging it just in case.

tcharding:
> * We really need to put some examples in the docs about how to turn a byteslice into a hex string.

Is this not clear enough (I added "arrays and slices" to try make it clearer):
```rust
//! let a = <[u8; 4]>::from_hex("deadbeef").expect("valid length and valid hex digits");
//! // We support `LowerHex` and `UpperHex` out of the box for `[u8]` arrays and slices.
//! println!("An array as lower hex: {:x}", a.as_hex());
//! // And for vecs since `Vec` derefs to byte slice.
//! println!("A vector as upper hex: {:X}", v.as_hex());
```

> * We should implement `Debug` and `Display` on `DisplayHex::Display` and have it behave the same as `LowerHex`. For `Debug` this is particularly important so that you can use `Formatter::debug_struct`.

I'm not following this?

> * I think we should rename `FromHex::Error` to `FromHex::Err` so that it matches the error variant for `FromStr`.

Done.

> * We need an example of how to implement`FromHex` on a type that wraps an array. It is confusing that the `from_byte_iter` method uses `HexToBytesError` but the trait needs to define its error type to use `HexToArrayError`.

We have that in `hexy.rs`, at the bottom of the file.

> * `fmt_hex_exact` needs examples of both forward and backward parsing

apoelstra:
> We have that in `hexy.rs`, at the bottom of the file.
> 

Defining a whole new structure when `DisplayArray` already exists and does 95% of the work? If this is the canonical/standard way to do hex serialaziotn then I'm not going to ever use this crate.

tcharding:
Do your comments on wrapping an array still stand after seeing #20?

apoelstra:
@tcharding it still involves manually creating a `DisplayALittleBitHexy` structure which appears to be identical to the existing `DisplayByteSlice`, so yes, my concerns still apply. It's fine to have an example of an advanced usecase but if a pure "wrapper around an array" involves rewriting tons of boilerplate, that's no good.

tcharding:
> but if a pure "wrapper around an array" involves rewriting tons of boilerplate, that's no good.

How about the `wrap_array_*` examples? Is that more boilerplate than you'd like. I rekon they are pretty minimal. The only problem I see is that it is not very clear (to me at least) why to favour one method over the other. Also even seeing both done I am unable to easily/clearly say which way is cleaner?

apoelstra:
@tcharding oh! Sorry, I didn't see those way at the bottom of the commit. Yes, those solve my concerns.

tcharding:
> Yes, those solve my concerns.

Sweet, I worked hard on those guys. I hope others find them when using this crate :)

tcharding:
From my position we are right to pull the trigger on this and release. Final boarding call for NACK's from @Kixunil.

apoelstra (approved):
utACK b2ac0dbe659d30e97319d3feb87aee685385e0a5

apoelstra (approved):
ACK b2ac0dbe659d30e97319d3feb87aee685385e0a5

apoelstra:
Actually lemme do one more "try to use this in rust-elements" run before merging. But agree that we've given Kix enough time to veto (more than enough time for a 0.1 release :P).

apoelstra:
Actually using this in rust-elements is blocked on actually using it in bitcoin_hashes (at least, my life woud be easier if we were to wait for bitcoin_hashes) since we would like to hex-serialize `sha256::Midstate` backward.

tcharding:
> Ok, the doccomment for fmt_hex_exact still needs an example of forward- and reverse-encoding hex.

Cool, I learned something while writing these. Came up with
```
/// Note that the bytes parameter is `IntoIterator` this means that if you would like to do some
/// manipulation to the byte array before formatting then you can. For example `bytes.iter().rev()`
/// to print the array backwards. 
```

I'll push this up because its trivial then look at the serde stuff separately.

PR #16, opened 2023-07-03
Implement suggestions after crate usage in rust-elements

tcharding:
Recently @apoelstra attempted to use this crate in `rust-elemements`, a few things came out in the wash (https://github.com/rust-bitcoin/hex-conservative/pull/15#issuecomment-1613578223)

Created #17 for the error rename.

This PR is a mess, changing to draft just to keep the discussion.

### DIFF
--- a/examples/hexy.rs
+++ b/examples/hexy.rs
@@ -60,9 +60,9 @@ impl fmt::UpperHex for Hexy {
 // And use a fixed size array to convert from hex.
 
 impl FromHex for Hexy {
-    type Error = HexToArrayError;
+    type Err = HexToArrayError;
 
-    fn from_byte_iter<I>(iter: I) -> Result<Self, Self::Error>
+    fn from_byte_iter<I>(iter: I) -> Result<Self, Self::Err>
     where
         I: Iterator<Item = Result<u8, HexToBytesError>> + ExactSizeIterator + DoubleEndedIterator,
     {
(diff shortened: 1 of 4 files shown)


apoelstra:
Deriving `Debug` in an example doesn't change anything about the API. The problem is that `DisplayByteSlice` and `DisplayArray` don't implement `Debug` or `Display`.

In general `DisplayHex::Display` should have `Debug` and `Display` bounds, and have documentation indicating that both of these should be equivalent to `LowerHex`.

tcharding:
> Deriving `Debug` in an example doesn't change anything about the API. The problem is that `DisplayByteSlice` and `DisplayArray` don't implement `Debug` or `Display`.

Excuse the changes I made yesterday, I was confused.

> In general `DisplayHex::Display` should have `Debug` and `Display` bounds, and have documentation indicating that both of these should be equivalent to `LowerHex`.

I think that is counter to the objective of the `DisplayHex` trait
```
/// Types that have a single, obvious text representation being hex should **not** implement this
/// trait and simply implement `Display` instead.
```

apoelstra:
> I think that is counter to the objective of the `DisplayHex` trait

Surely `DisplayBytesSlice` and `DisplayArray` have a "single, obvious text representation"?

tcharding:
Ok, I spent literally hours on this crate today. FTR this code is way too complicated for such a trivial task as converting bytes to/from hex. Anyway, whinge over. I'll push a few PRs with what I came up with.

tcharding:
I think this can be closed now, nothing here needs action still.

PR #17, opened 2023-07-03
Use Err for the associated error type of FromHex

tcharding:
The `FromHex` trait parses a hex string, as such it is quite like `FromStr` - it makes sense for the associated error to be named the same `Err` instead of `Error` (is it is in `TryFrom` and other places).


apoelstra (approved):
ACK 75c6c69dbcd2cb47b1d11bcc41923eae93687a58

apoelstra:
I guess I will manually close this since Github can't detect the merge..

PR #20, opened 2023-07-04
Add examples

tcharding:
Excludes patch 1 (#21).

Add additional examples. Patches 2-4 are preparatory cleanup.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,3 +31,12 @@ core2 = { version = "0.3.2", default_features = false, optional = true }
 
 [[example]]
 name = "hexy"
+
+[[example]]
+name = "custom"
+
+[[example]]
+name = "wrap_array_display_hex_trait"
+
+[[example]]
+name = "wrap_array_fmt_traits"
--- a/examples/custom.rs
+++ b/examples/custom.rs
@@ -1,0 +1,161 @@
+//! Demonstrate hexadecimal encoding and decoding for a type where hex is not the natural hex representation but the type can still be encoded/decoded to/from hex.
+//!
+//! For a type where hex is the natural representation see `./hexy.rs`.
+//! To wrap an array see the `./wrap_array_*` examples.
+
+use core::fmt;
+use core::str::FromStr;
+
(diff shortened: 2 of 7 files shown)


apoelstra:
I"m tempted to blanket-impl both `Debug` and `Display` on all implementors of this trait, passing through to `LowerHex`. Though I could imagine that causing issues down the line.

apoelstra (approved):
ACK b97b21aed7090a0c5666987caef9f4482557fa57

apoelstra:
Will hold off on merging and give @Kixunil 72 hours to chime in. Though I really think that adding `Debug` impls on stuff is non-controversial, especially as it's a non-breaking change to modify their output.

tcharding:
I pulled the trait bounds stuff out into #21 because I want to play with these examples a bit more.

tcharding:
> I"m tempted to blanket-impl both Debug and Display on all implementors of this trait, passing through to LowerHex. 

That is definitely something that can be done in a future version, right? v0.1.0 is ok like this I rekon.

tcharding:
I'm not clear on how to move forward on this now. I'll re-read the thread tomorrow but if you want to please say.

tcharding:
> I"m tempted to blanket-impl both Debug and Display on all implementors of this trait, passing through to LowerHex.

On further thought, and after doing the wrap array examples this might be quite useful to differentiate between the two ways to use the lib a) impl fmt traits (and you do not get the default/display impls so you can control them yoursef) or implement the trait and get default/display impls.

> Though I could imagine that causing issues down the line.

Is this still a concern after my comment above?

tcharding:
> I"m tempted to blanket-impl both Debug and Display on all implementors of this trait, passing through to LowerHex.

This does not apply now that we have the two `wrap_array_*` examples showing the two ways to use the display module, right?

apoelstra:
> This does not apply now that we have the two `wrap_array_*` examples showing the two ways to use the display module, right?

No, it was just an idle thought.

apoelstra (approved):
ACK 96fd379d08f0578ad22fe0564557caad840814f7

PR #22, opened 2023-07-13
Add more documentation to the fmt_hex_exact macro

tcharding:
The `fmt_hex_exact` macro takes bytes as `IntoIterator`, this makes it quite flexible. Add more documentation to the macro, giving an example of formatting an array in reverse.


apoelstra (approved):
ACK 98463356f1398d42a3bfb57801777d39f8161e17

Issue #23, opened 2023-07-13
Width/precision not implemented for bytes/array as_hex


apoelstra:
You cannot do `format!("{:8}", bytes.as_hex())` and get an 8-character prefix the way you used to be able to do with bitcoin_hashes::hex.

apoelstra:
It looks like we actually lost this ability in bitcoin_hashes 0.12 even when displaying hashes. It existed in bitcoin_hashes 0.11.

That's what we get for releasing without any support for hex encoding .... nobody even tried to use this.

PR #25, opened 2023-07-14
Improve formatting

tcharding:
Currently the buffered encoder does not handle padding, this is because it writes in chunks so cannot.
    
We would like to support padding; observe that it is unlikely that anyone will pad a hex string that is larger than one chunk of the buffered encoder (1024 hex characters) so we can just pad for slices shorter than this.

Patch 2 adds a unit test that can be move in front to verify this did not work and now does.

Fix: #23

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -4,6 +4,26 @@
 //!
 //! This module provides a trait for displaying things as hex as well as an implementation for
 //! `&[u8]`.
+//!
+//! For arrays and slices we support padding and precision for length < 512 bytes.
+//!
+//! # Examples
+//!
+//! ```
+//! use hex_conservative::DisplayHex;
+//!
+//! // Display as hex.
+//! let v = vec![0xde, 0xad, 0xbe, 0xef];
+//! assert_eq!(format!("{}", v.as_hex()), "deadbeef");
+//!
+//! // Get the most significant bytes.
+//! let v = vec![0x01, 0x23, 0x45, 0x67];
+//! assert_eq!(format!("{0:.4}", v.as_hex()), "0123");
+//!
+//! // Padding with zeros
+//! let v = vec![0xab; 2];
+//! assert_eq!(format!("{:0>8}", v.as_hex()), "0000abab");
+//!```
 
 use core::borrow::Borrow;
 use core::fmt;
@@ -136,6 +156,13 @@ impl<'a> DisplayByteSlice<'a> {
     fn display(&self, f: &mut fmt::Formatter, case: Case) -> fmt::Result {
         let mut buf = [0u8; 1024];
         let mut encoder = BufEncoder::new(&mut buf);
+
+        // Its unlikely that someone will want special formatting for a hex string that
+        // is over 1024 characters so just handle padding for short slices.
+        if self.bytes.len() < 512 {
+            encoder.put_bytes(self.bytes, case);
+            return f.pad(encoder.as_str());
+        }
 
         let mut chunks = self.bytes.chunks_exact(512);
         for chunk in &mut chunks {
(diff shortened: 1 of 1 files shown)


apoelstra:
Can you add a unit test which checks what happens when you try to encode a string longer than 1024 characters? What if you try to set padding more than 1024 characters?

We also need to support the `precision` option of `Formatter`, which is arguably more important.

tcharding:
I've used `pad` now instead of `pad_integral` and (after learning a bit more about formatting strings) I've added rustdoc examples to the `display` module and a unit test verifying no padding on long arrays. Does that cover everything you had in mind?

apoelstra:
Ok, I'm happy with this. I guess we should file a bug to support padding for longer strings, since it wouldn't be *too* hard to support it, it's just a PITA and we want to get this out.

But we still need to make `{:8}` truncate the hex string to 8 characters, so that you can output truncated hashes.

tcharding:
I'm confused, the docs for `Formatter::pad` say 

> precision - the maximum length to emit, the string is truncated if it is longer than this length

But neither of the following two get truncated 
```rust

    #[test]
    fn test() {
        let x = 1000;
        println!("{:2}", x);
        let s = "1000";
        println!("{:2}", s);
    }
```

apoelstra:
@tcharding your two examples are numbers, not strings.

tcharding:
To clarify I'm understanding what behaviour you expect, you want the following to things to happen when using `{:8}`, correct?
```rust
    #[test]
    fn precision() {
        let v = vec![0x12, 0x34, 0x56, 0x78, 0x9a, 0xbc, 0xde, 0x0f];
        let got = format!("{:.8}", v.as_hex());
        let want = "12345678";
        assert_eq!(got, want);

        let v = vec![0x12, 0x34];
        let got = format!("{:0>8}", v.as_hex());
        let want = "00001234";
        assert_eq!(got, want);
    }
```

tcharding:
ooooo, the problem with my original example is `{:2}` is wrong - its supposed to be `{:.2}`. So at the moment we have the same behaviour as a string
```rust

    #[test]
    fn tobin() {
        let x = 1000;
        println!("int truncate 2 chars: {:.2}", x);
        let s = "1000";
        println!("integer str truncate 2 chars: {:.2}", s);
        let s = "abcd";
        println!("str truncate 2 chars: {:.2}", s);

        let pi: f32 = 3.1415926536897932384626;
        println!(" (float) pi is {:.32}", pi);
        // Truncates to 8 decimal places.
        println!(" (float) pi is {:.8}", pi);

        let pi = "3.1415926536897932384626";
        println!("(string) pi is {:.32}", pi);
        // Truncates to 8 characters.
        println!("(string) pi is {:.8}", pi);

        let v = vec![0x12, 0x34, 0x56, 0x78, 0x9a, 0xbc, 0xde, 0x0f];
        println!("hex too long prec 8 : {:.8}", v.as_hex());

        let v = vec![0x12, 0x34];
        println!("hex too short prec 8: {:.8}", v.as_hex());
    }
```

apoelstra:
Ok, checked that this matches the behavior for `String`. It does. Looks good!

apoelstra:
@tcharding it looks like you've added an example in doccomments but I don't see any code or tests for the precision behavior.

apoelstra:
Looks like the logic is already there. Ok, I guess we just need some tests.

tcharding:
> @tcharding it looks like you've added an example in doccomments but I don't see any code or tests for the precision behavior.

More than whats there already?
```rust
//! // Get the most significant bytes.
//! let v = vec![0x01, 0x23, 0x45, 0x67];
//! assert_eq!(format!("{0:.4}", v.as_hex()), "0123");
```

apoelstra:
@tcharding I guess, the doccomment is actually a test. But I meant an explicit unit test.

tcharding:
Oh I intentionally removed the unit test when I added the rustdoc test, I've often done this. Do you think there is value in having the same functionality tested in rustdoc as well as a unit test? One thing I could think is that the rustdoc could get changed/removed without the unit test being put back so we'd loose coverage.

apoelstra:
I guess there's not really additional value. But it helps during review to just look at the unit tests and see that everything's tested. The unit tests are also a better source of documentation than doccomments (ironic, I know) because there is a lot of different usage in one place alongside expected outputs.

tcharding:
Ok, so rustdoc is good for basic docs and regression testing the public API and unit tests for full coverage, both as tests and as docs - nice. I'll do a PR.

tcharding:
Ha, I still had to think what would the behaviour would be when I wrote the test, and I wrote this code within the last two weeks - proving your point that unit tests are better docs :)

https://github.com/rust-bitcoin/hex-conservative/pull/27

PR #26, opened 2023-07-14
Bump version to 0.1.1

tcharding:
- Patch 1 is #25, required for this release.
- Patch 2 bumps the version and adds changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.1.1 - 2023-07-19
+
+- [Add `test_hex_unwrap`](https://github.com/rust-bitcoin/hex-conservative/pull/24) hex parsing macro for test usage.
+- [Improve formatting](https://github.com/rust-bitcoin/hex-conservative/pull/25) hex for bytes slices e.g., support padding.
+
 # 0.1.0 - 2023-06-20 Initial Release
 
 - [Import](https://github.com/rust-bitcoin/hex-conservative/pull/1) code from the `bitcoin_hashes` and `bitcoin-internals` crates.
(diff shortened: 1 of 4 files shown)


apoelstra:
I think we could probably get away with 0.1.1. But up to you.

tcharding:
Number go up!

tcharding:
Not sure why I have been so confused for the last few years, this seems straight forward enough

> Initial development releases starting with “0.y.z” can treat changes in “y” as a major release, and “z” as a minor release.

https://doc.rust-lang.org/cargo/reference/semver.html

Changed to use v0.1.1 - note the branch name is now incorrect.

apoelstra:
> Not sure why I have been so confused for the last few years, this seems straight forward enough

I am really fascinated to know what made this click for you. I feel like Kix and I have said that exact sentence before.

In any case, ack using 1.1 :)

tcharding:
> I am really fascinated to know what made this click for you. I feel like Kix and I have said that exact sentence before.

If you ever work out how my brain works I'll be first in line to hear it. 90% of the time it feels like a well oiled machine and then it has these gaping holes in it, I am constantly baffled by the workings of the human mind.

tcharding:
I think it has something to do with "tunnel vision", so slightly unrelated, but yesterday I was at an area and talking about a kids park, I pointed to where I thought it was. A few minutes later, after walking some more, I looked over again and the park was exactly next to where I had pointed and I had not seen it. The visual system is not the whole brain but these sorts of fails highlight similar workings I rekon. I get the sense that our brains actually do a lot less than we think (i.e., process way less visual inputs in this case) but in a way that makes them seem like they are doing alot.

apoelstra (approved):
ACK 42f3cdad67fd7f3a07058c5e231aaf4b12d453aa

apoelstra:
Tagged and published.

PR #28, opened 2023-07-25
Rename the library to hex

tcharding:
Currently this crate is used using either `use hex-conservative::Foo` or using the `package` trick in the dependency specification to shorten it to `hex`. However the manifest provides a `lib` section for library renaming, we can use this so that the lib can be imported as `hex`.

With this applied the dependency specification becomes `hex = X.Y.Z` and the usage becomes `use hex::DisplayHex`.

This can be seen to work by checking the `fuzz` crate.

@stevenroose argued for this ages ago but at the time I did not manage to process the argument and thought this did not work - turns out he was right.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,9 @@ readme = "README.md"
 edition = "2018"
 exclude = ["tests", "contrib"]
 
+[lib]
+name = "hex"
+
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
--- a/examples/custom.rs
+++ b/examples/custom.rs
@@ -6,7 +6,7 @@
 use core::fmt;
 use core::str::FromStr;
 
-use hex_conservative::{DisplayHex, FromHex, HexToArrayError, HexToBytesError};
+use hex::{DisplayHex, FromHex, HexToArrayError, HexToBytesError};
 
 fn main() {
     let s = "deadbeefcafebabedeadbeefcafebabedeadbeefcafebabedeadbeefcafebabe";
--- a/examples/hexy.rs
+++ b/examples/hexy.rs
@@ -6,9 +6,7 @@
 use std::fmt;
 use std::str::FromStr;
 
-use hex_conservative::{
-    fmt_hex_exact, Case, DisplayHex, FromHex, HexToArrayError, HexToBytesError,
-};
+use hex::{fmt_hex_exact, Case, DisplayHex, FromHex, HexToArrayError, HexToBytesError};
 
 fn main() {
     let s = "deadbeefcafebabedeadbeefcafebabedeadbeefcafebabedeadbeefcafebabe";
--- a/examples/wrap_array_display_hex_trait.rs
+++ b/examples/wrap_array_display_hex_trait.rs
@@ -4,8 +4,8 @@
 //!
 //! For an example using the standard library `fmt` traits see `./wrap_array_fmt_traits.rs`.
 
-use hex_conservative::display::DisplayArray;
-use hex_conservative::{DisplayHex, FromHex, HexToArrayError, HexToBytesError};
+use hex::display::DisplayArray;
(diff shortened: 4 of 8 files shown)


tcharding:
We will need to release v0.2.0 if/when this merges. I've put the [usage PR](https://github.com/rust-bitcoin/rust-bitcoin/pull/1833) to draft awaiting resolution of this PR.

apoelstra:
How is this used? Why won't it conflict with the existing `hex` crate on crates.io?

apoelstra:
It looks like you put `hex-conservative` into Cargo.toml and then magically a `hex` symbol appears in your crate, because of the manifest for `hex-conservative` and not because of anything actually in your codebase? This is like a wildcard import on steroids. To find the symbol you need to download and check every single dependency.

tcharding:
@stevenroose assures me that this works and it can be seen in action by looking at
- https://github.com/DaGenix/rust-crypto/blob/master/Cargo.toml
- https://github.com/RustCrypto/traits/blob/master/crypto/Cargo.toml

It does make the rustdocs online a bit confusing because both use the name `crypto`.

tcharding:
oh and if you put `use hex_conservative::Foo` you get the error message
```
error[E0432]: unresolved import `hex_conservative`
 --> fuzz/fuzz_targets/hex.rs:4:5
  |
4 | use hex_conservative::{fmt_hex_exact, Case, FromHex, HexToArrayError, HexToBytesError};
  |     ^^^^^^^^^^^^^^^^ use of undeclared crate or module `hex_conservative`
```

And have no way of knowing you must use `hex` without looking at the manifest - that is rubbish.

stevenroose:
Yeah so your Rust library can have a different name than the crate. The crate just contains the library but it can also contain various binaries.

So in the manifest file, the `[lib] name = "hex"` section defines the name of the library.

So yes it's true that if you use both our hex and another crate that has the hex library name, it will conflict and you might have to do some fiddling to get that working. It's like the messing we do with `actual-serde` and stuff, even though that's with crate names and not library names.

I do actually think the way you have to rename the libraries is actually confusing and badly designed from cargo.

But if you have f.e. dependencies `crypto` and `rust-crypto`, you'll get this error:
```
    Checking rusttest v0.1.0 (/home/steven/tmp/rusttest)
error[E0465]: multiple rmeta candidates for `crypto` found
 --> src/main.rs:2:5
  |
2 | use crypto::*;
  |     ^^^^^^
  |
note: candidate #1: /var/cargo-target/debug/deps/libcrypto-b4f05945fe115be8.rmeta
 --> src/main.rs:2:5
  |
2 | use crypto::*;
  |     ^^^^^^
note: candidate #2: /var/cargo-target/debug/deps/libcrypto-873be0620c80c494.rmeta
 --> src/main.rs:2:5
  |
2 | use crypto::*;
  |     ^^^^^^

error: could not compile `rusttest` due to previous error
```

The solution is to do
```toml
[dependencies]
crypto = "0.5.1"
crypto2 = { version = "0.2.36", package="rust-crypto"}
```

So here it's very clearly conflating the package, library and crate concepts, but well. Cargo being cargo.

stevenroose:
But the main point is, this is a hex library, not a hex_conservative library. Developers that want two hex libraries are silly. Won't we export this as `bitcoin::hex` or so anyway because we have to kinda?

tcharding:
> Won't we export this as bitcoin::hex or so anyway because we have to kinda?

This point mutes this whole PR because who is going to use this crate anyways other than by way of either `hashes` or `bitcoin` (both will have re-exports).

apoelstra:
>This point mutes this whole PR because who is going to use this crate anyways other than by way of either hashes or bitcoin (both will have re-exports).

All the Elements crates use it directly since `bitcoin` is feature-gated. Using it through `hashes` is a neat workaround. In any case, we re-export it there as `hex`. So I think the current situation is working fine, and this business with `[lib] name = "hex"` is clever but IMHO too clever.

PR #30, opened 2023-08-10
Inline public methods

tcharding:
Audit the whole crate and add `#[inline]` as needed.


apoelstra (approved):
ACK 501a957c34e4807b0b0ecbde9c6d3c8e49bdfbfa

PR #31, opened 2023-10-25
Audit error types

tcharding:
Audit the crate error types (all two of them) and ensure they follow the "standard" we are starting to settle on here in the rust-bitcoin org.

### DIFF
--- a/examples/custom.rs
+++ b/examples/custom.rs
@@ -117,7 +117,7 @@ impl<'a> fmt::UpperHex for DisplayALittleBitHexy<'a> {
 }
 
 /// Example Error.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum Error {
     /// Conversion error while parsing hex string.
     Conversion(HexToBytesError),
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -147,6 +147,7 @@ impl std::error::Error for HexToArrayError {
 }
 
 impl From<HexToBytesError> for HexToArrayError {
+    #[inline]
     fn from(e: HexToBytesError) -> Self { Self::Conversion(e) }
 }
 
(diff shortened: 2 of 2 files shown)


examples/custom.rs
```diff
@@ -117,7 +117,7 @@ impl<'a> fmt::UpperHex for DisplayALittleBitHexy<'a> {
 }
 
 /// Example Error.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Error {
     /// Conversion error while parsing hex string.
     Conversion(HexToBytesError),
```
Kixunil, line 123: I don't see `InvalidStringFormat` below used anywhere is it still needed?

Kixunil:
As for ordering `Copy`/`Clone` do we have a standard order now? For me `Copy` is mentally more important since `Clone` is implied, so previous order makes more sense (and I write it int that order) but I guess this is too much bikeshedding.

tcharding:
> As for ordering Copy/Clone do we have a standard order now? For me Copy is mentally more important since Clone is implied, so previous order makes more sense (and I write it int that order) but I guess this is too much bikeshedding.

Drats, I wish you were here a few months ago, I've made this change in so many places. I agree with you (on the ordering not the too much bikeshedding :) - FWIW when trying to come up with an order I went with `Copy` depends on `Clone` being there so put it after, I like yours better though.

You owe me one, for the shitstorm I'm going to get when I patch the whole org one Friday afternoon with this change.

apoelstra:
>You owe me one, for the shitstorm I'm going to get when I patch the whole org one Friday afternoon with this change.

Lol! FWIW I really don't think it matters as long as they're beside each other.

tcharding:
This is a text book case of "just have one way of doing it and do it like that ... for ever". Maybe I'm too stuck on this principle and too easily triggered, we all have weaknesses eh.

tcharding:
Needs #33 to get past CI

Kixunil:
Yeah, not worth reverting, I was just curious about it.

Kixunil (approved):
ACK 9d851d4ff87908c102e1a45c552c978f6647e337

apoelstra (approved):
ACK 9d851d4ff87908c102e1a45c552c978f6647e337

PR #34, opened 2023-10-26
Check for changes to the public API

tcharding:
We would like to get to a stage where we can commit to the public API. To help us achieve this add a script that generates the public API and checks it against three committed files, one for each feature set: no features, alloc, std.

The idea is that with this applied any PR that changes the public API should include a final patch that is just the changes to the api/*.txt files, that way reviewers can discuss the changes without even needing to look at the code, quickly giving concept ACK/NACKs. We also run the script in CI to make sure we have not accidentally changed the public API so that we can be confident that don't break semver during releases. The script can also be used to diff between two release versions to get a complete list of API changes, useful for writing release notes and for users upgrading.

There is a development burden involved if we apply this patch.

### DIFF
--- a/contrib/check-for-api-changes.sh
+++ b/contrib/check-for-api-changes.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+#
+# Checks the public API of crates, exits with non-zero if there are currently
+# changes to the public API not already committed to in the various api/*.txt
+# files.
+
+set -e
+
+export RUSTDOCFLAGS='-A rustdoc::broken-intra-doc-links'
+REPO_DIR=$(git rev-parse --show-toplevel)
+API_DIR="$REPO_DIR/api"
+CMD="cargo +nightly public-api --simplified"
+# `sort -n -u` doesn't work for some reason.
+SORT="sort --numeric-sort"
+
+# cargo public-api uses nightly so the toolchain must be available.
+if ! cargo +nightly --version > /dev/null; then
+    echo "script requires a nightly toolchain to be installed (possibly >= nightly-2023-05-24)" >&2
+    exit 1
+fi
+
+pushd "$REPO_DIR" > /dev/null
+$CMD --no-default-features | $SORT | uniq > "$API_DIR/no-default-features.txt"
+$CMD | $SORT > "$API_DIR/default-features.txt"
+$CMD --no-default-features --features=alloc | $SORT | uniq > "$API_DIR/alloc.txt"
+$CMD --no-default-features --features=core2 | $SORT | uniq > "$API_DIR/core2.txt"
+$CMD --all-features | $SORT | uniq > "$API_DIR/all-features.txt"
+
+if [[ $(git status --porcelain api) ]]; then
+    echo "You have introduced changes to the public API, commit the changes to api/ currently in your working directory" >&2
+    popd > /dev/null
+    exit 1
+else
+    echo "No changes to the current public API"
+    popd > /dev/null
+fi
+
+exit 0
+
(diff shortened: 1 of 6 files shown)


tcharding:
This includes one for "core2" and one for "alloc", separately because its easy to mess up feature gates with these two.

tcharding:
@Kixunil are you good with this?

contrib/check-for-api-changes.sh
```diff
@@ -0,0 +1,32 @@
...
+if ! cargo +nightly --version > /dev/null; then
+    echo "script requires a nightly toolchain to be installed (possibly >= nightly-2023-05-24)" >&2
+    exit 1
+fi
+
+pushd "$REPO_DIR" > /dev/null
+$CMD --no-default-features | sort --unique > "$API_DIR/no-default-features.txt"
+$CMD | sort --unique > "$API_DIR/default-features.txt"
+$CMD --no-default-features --features=alloc | sort --unique > "$API_DIR/alloc.txt"
+$CMD --no-default-features --features=core2 | sort --unique > "$API_DIR/core2.txt"
```
Kixunil, line 24: Could you try sorting with `-n`? It would read better if the array types were sorted naturally.

contrib/check-for-api-changes.sh
```diff
@@ -0,0 +1,32 @@
...
+$CMD --all-features | sort --unique > "$API_DIR/all-features.txt"
+
+if [[ $(git status --porcelain api) ]]; then
+    echo "You have introduced changes to the public API, commit the changes to api/ currently in your working directory" >&2
```
Kixunil, line 28: Should this exit with non-zero?

Kixunil (commented):
Concept ACK.

contrib/check-for-api-changes.sh
tcharding: Using `-n` (or `-h`) causes the script to produce no output.

contrib/check-for-api-changes.sh
tcharding: Yes, nice. Done.

apoelstra (approved):
ACK 5d7eaf5cc59fc5ae46d3107c50b5b6a41257a8b9

contrib/check-for-api-changes.sh
Kixunil: WTF, that's some weird interaction between `-n` and `--unique`. You can fix it by writing `sort -n | uniq`

tcharding: So I think we have pushed bash as far as it can reasonably go, bash can't easily sort the following 4 lines as we want
```
impl<'a> hex_conservative::display::DisplayHex for &'a [u8; 9]
impl<'a> hex_conservative::display::DisplayHex for &'a [u8; 8]
pub type [u8; 128]::Err = hex_conservative::HexToArrayError
pub type [u8; 10]::Err = hex_conservative::HexToArrayError
```

Are you guys ok with a perl script to get the order to be:
```
impl<'a> hex_conservative::display::DisplayHex for &'a [u8; 8]
impl<'a> hex_conservative::display::DisplayHex for &'a [u8; 9]
pub type [u8; 10]::Err = hex_conservative::HexToArrayError
pub type [u8; 128]::Err = hex_conservative::HexToArrayError
```

Is there a better tool than perl for this? (FTR neither my perl nor awk are good enough to quickly write this but since we are going to use this script everywhere I can allocate some time to it.)

tcharding:
Added piping through `uniq` which is clearly better. Whether or not we should improve the sort order I don't know, I can definitely put an hour or so into it.

contrib/check-for-api-changes.sh
apoelstra: Either awk or perl is fine by me. I can read, but not really write, those languages.

Kixunil: Not a fan of perl but if it's the simplest choice for you I could tolerate it.

Kixunil (approved):
ACK b45bd6ef2c785073b854f4c19c21281028a41fd8

PR #36, opened 2023-10-26
Add a prelude module

tcharding:
We really only have a few traits and types that users need to worry about when using this crate. Feels like `use hex::prelude::*` is ergonomic and unsurprising.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,7 +10,7 @@
 //! // In your manifest use the `package` key to improve import ergonomics.
 //! // hex = { package = "hex-conservative", version = "*" }
 //! # use hex_conservative as hex; // No need for this if using `package` as above.
-//! use hex::{DisplayHex, FromHex};
+//! use hex::prelude::*;
 //!
 //! // Decode an arbitrary length hex string into a vector.
 //! let v = Vec::from_hex("deadbeef").expect("valid hex digits");
@@ -49,10 +49,10 @@ mod error;
 mod iter;
 pub mod parse;
 
-/// Reexports of extension traits.
-pub mod exts {
-    pub use super::display::DisplayHex;
-    pub use super::parse::FromHex;
+/// Re-exports of the common crate traits.
+pub mod prelude {
+    #[doc(inline)]
+    pub use crate::{display::DisplayHex, parse::FromHex};
 }
 
 #[rustfmt::skip]                // Keep public re-exports separate.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -49,10 +49,13 @@ mod error;
...
-pub mod exts {
-    pub use super::display::DisplayHex;
-    pub use super::parse::FromHex;
+/// Reexports of the common crate types and traits.
+pub mod prelude {
+    #[doc(inline)]
+    pub use crate::{
+        display::DisplayHex,
+        parse::{FromHex, HexToArrayError, HexToBytesError},
+    };
```
apoelstra, line 58: In 326aad38caede931157b7e5f9d6ba6b650319670:

At first I didn't like this, but I see that everything in the `exts` module is re-exported at the crate root. So it seems like `exts` was already supposed to be a sort of `prelude` module, and I agree it would be better if it were just named `prelude`.

apoelstra (approved):
ACK 326aad38caede931157b7e5f9d6ba6b650319670

src/lib.rs
Kixunil: Kind of, it was meant to only contain extension traits, preludes often contain other things. But I guess it's better to use `prelude` anyway.

src/lib.rs
```diff
@@ -49,10 +49,13 @@ mod error;
...
-/// Reexports of extension traits.
-pub mod exts {
-    pub use super::display::DisplayHex;
-    pub use super::parse::FromHex;
+/// Reexports of the common crate types and traits.
+pub mod prelude {
+    #[doc(inline)]
+    pub use crate::{
+        display::DisplayHex,
+        parse::{FromHex, HexToArrayError, HexToBytesError},
```
Kixunil, line 57: Do we really need the errors in prelude? I think most users will not use them unless they are making a library in which case they probably shouldn't use prelude at all.

tcharding: Ah yes, I agree. Will remove.

tcharding: Removeing the error types makes this PR a straight re-name of `exts` to `prelude`. I re-wrote the git log to say as much but left the PR descriptions as it was.

Kixunil (approved):
ACK 539202539b5f7ab9572a874decc93f8d0b258600

apoelstra (approved):
ACK 539202539b5f7ab9572a874decc93f8d0b258600

Issue #13, opened 2023-05-18
Error handling


tcharding:
Is it ok that `HexToBytesError` does not use `non_exhaustive`. I think it is but could I get a few more eyes on it please.

apoelstra:
Yeah, I think this is reasonable. There are only so many ways that hex decoding could go wrong and these variants all contain plausibly useful information for users to match on.

tcharding:
If #31 merges I believe we can close this.

Kixunil:
Yeah, it's good, I just want to wrap each variant in a struct. Especially `Char` one where I want to support reporting multiple chars eventually (with feature flags and such).

Kixunil:
If the intention is to release this ASAP as 0.x, then I'm fine. But for 1.0 I want the internals hidden. I specifically know what I want to add eventually (spans), so I see no reason to not protect the stability.

However, I'm considering allowing in the future to be able to report both uneven length and wrong characters, which would imply hiding whole error and then having a method to get the first error (which is always available) and a method to iterate over the errors when `multi-span` feature is enabled.

apoelstra:
Yes, for 1.0 we definite want to hide the errors properly and add as much extra information as we can.

@tcharding I wouldn't mind doing some small things in order to get a quick release out, but I wonder if we actually need a quick release here? This crate is functionally complete and the remaining TODOs are API improvements for 1.0, so maybe we should just grind away on that and not bother releasing til we think we're close to done?

tcharding:
Oh I wasn't pushing for another 0.x release, was just referring to the eventual 1.0 release - which I am pushing for :)

tcharding:
Oh boy, I was just working on this and I realized that even wrapping the errors in structs does not save us from having to do a major version bump when we add information because doing so almost certainly changes the `Display` implementation which is part of the public API (I read this somewhere recently).

Kixunil:
@tcharding I'm not convinced `Display` *implementation* *has to* be part of public API, especially for errors but the solution is simply to not put that information into `Display` and provide a different way to access it.

For spans I would absolutely *not* put them into `Display` anyway (unless they were "on line x, character y" but we won't have those).

tcharding:
> but the solution is simply to not put that information into Display and provide a different way to access it.

Oh, that is a good point. Did not think of that. Cheers.

tcharding:
Closing in favour of #42

PR #40, opened 2023-10-27
Add `InvalidLengthError`

tcharding:
Draft for discussion. This wraps the internals of the `HexToArrayError` as suggested in #13.

Note: Please don't take this as argumentative, the purpose is to get right to the bottom of things so once we start executing whatever we decide on its done right. Remember, we are setting precedent for all the other crates in the `rust-bitcoin` org.

I personally think this is going too far for a two reasons, primarily (1)

1. This does not save us from having to do a major release when we add new data to the error types because most likely that will lead to a change in the `Display` output which is a major breaking change.
2. We already have quite a lot of error boilerplate, admittedly this is write-once code but still it is there, its still a liability. This new wrapping style introduces loads more boilerplate, which means loads of work, which likely no one else is going to do except me - and its boring as bat shit. If we are going to go for this I'd like everyone to be _really_ confident its worth it. Also everytime we have to make a tiny change to some stylistic error thing it has to be done in many places, and I never get it right first time.

### DIFF
--- a/examples/custom.rs
+++ b/examples/custom.rs
@@ -117,7 +117,7 @@ impl<'a> fmt::UpperHex for DisplayALittleBitHexy<'a> {
 }
 
 /// Example Error.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Error {
     /// Conversion error while parsing hex string.
     Conversion(HexToBytesError),
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -7,6 +7,7 @@ use core::{fmt, str};
 #[cfg(all(feature = "alloc", not(feature = "std")))]
 use crate::alloc::vec::Vec;
 use crate::iter::HexToBytesIter;
+use crate::write_err;
 
 /// Trait for objects that can be deserialized from hex strings.
 pub trait FromHex: Sized {
@@ -86,7 +87,9 @@ macro_rules! impl_fromhex_array {
                     }
                     Ok(ret)
                 } else {
-                    Err(HexToArrayError::InvalidLength(2 * $len, 2 * iter.len()))
+                    let expected = 2 * $len;
+                    let got = 2 * iter.len();
+                    Err(InvalidLengthError { expected, got }.into())
                 }
             }
         }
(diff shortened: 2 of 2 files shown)


Kixunil:
> 1

Adressed in #13 TL;DR: not an issue

> 2

I believe it's worth it and we don't really have any alternative other than wrapping whole error and disclosing all the details later or just getting everything right the first time.

I'm perfectly willing to write that boilerplate, I'm just currently doing some more urgent things unrelated to this crate.

Kixunil:
Oh, actually we may have to wrap the top-level error as well to support multiple errors (unless we want to split the API, which I'd prefer not to).

Roughly:

```rust
// impl Error such that only first is considered and it's transparent
pub struct Error {
    first: HexError,
    // We could also do clever things with Vec to make the error type smaller but that's for another day
    // Note that we don't do internal enforcing of sanity (preventing duplicates) because it'd make the API awful.
    #[cfg(feature = "multi-error")]
    second: Option<HexError>,
}

impl Error {
    pub fn first(&self) -> &HexError {
        &self.first
    }

    pub fn into_first(self) -> HexError {
        self.first
    }

    pub fn iter(&self) -> impl Iterator<Item = &HexError> {
        core::iter::once(&self.first)
            #[cfg(feature = "multi-error")]
            .chain(self.second.as_ref())
    }
    // into_iter as well

    pub fn into_vec(self) -> Vec<HexError> {
        #[cfg(feature = "multi-error")]
        match self.second {
            Some(second) => vec![self.first, second],
            None => vec![self.first],
        }
        #[cfg(not(feature = "multi-error"))]
        vec![self.first]
    }
}

// impl Error
pub enum HexError {
    OddLength(OddLengthError),
    InvalidChars(InvalidChars),
}
```

tcharding:
> I'm perfectly willing to write that boilerplate, I'm just currently doing some more urgent things unrelated to this crate.

Ok, I'm going to leave the error handling in this crate to you. I'm guessing the snippet above goes beyond what will need doing in other crates since you have an idea of what is going to be done here in future, where as I was just trying to come up with a precedent to apply to all the other crates (initially `rust-bech32` and `hashes` since they, along with this one, are likely the first to stabalise).

tcharding:
Closing in favour of #42

PR #41, opened 2023-10-27
Remove arbitrary padding limit

Kixunil:
When formatting a slice as hex an arbitrary limit of 512 chars was applied to decide whether to pad it. This was silent and surprising so this commit fixes it by removing the limit. It implements padding manually with moderate optimizations.

(The limit was introduced in #25 )

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -154,24 +154,82 @@ pub struct DisplayByteSlice<'a> {
 
 impl<'a> DisplayByteSlice<'a> {
     fn display(&self, f: &mut fmt::Formatter, case: Case) -> fmt::Result {
+        use fmt::Write;
+        // There are at least two optimizations left:
+        //
+        // * Reusing the buffer (encoder) which may decrease the number of virtual calls
+        // * Not recursing, avoiding another 1024B allocation and zeroing
+        //
+        // This would complicate the code so I was too lazy to do them but feel free to send a PR!
+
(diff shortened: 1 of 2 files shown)


src/display.rs
```diff
@@ -154,24 +154,82 @@ pub struct DisplayByteSlice<'a> {
...
+            };
+
+            if string_len < width {
+                let (left, right) = match f.align().unwrap_or(fmt::Alignment::Left) {
+                    fmt::Alignment::Left => (0, width - string_len),
+                    fmt::Alignment::Right => (width - string_len, 0),
+                    fmt::Alignment::Center =>
+                        ((width - string_len) / 2, (width - string_len + 1) / 2),
+                };
+                // Avoid division by zero and optimize for common case.
```
apoelstra, line 181: In 8569ee8e55da4ecb099c68dca7eceb9ebae619ba:

Here and below I don't think there is any actual division by 0. There would be "dividing 0 by something" but this is harmless.

Kixunil: `chunk_len` is 0 if the input to `put_filler` is 0. I actually got panics for that, unfortunately I wasn't smart enough to see it first. :)

apoelstra: Lol, me neither apparently :)

apoelstra (approved):
ACK 8569ee8e55da4ecb099c68dca7eceb9ebae619ba

Issue #42, opened 2023-10-30
Totally conceal error internals


tcharding:
Before stabalizing we want to totally conceal the internals of all error types in this crate. 

This was discussed in:
- #13 
- #40

PR #44, opened 2023-11-06
Hide all error internals

tcharding:
Before we do v1.0 we want to hide all the error internals so that we can modify/improve them without breaking the public API.

Please pay particular attention to the `Display` strings because we they are considered part of the public API so we don't want to change them willy-nilly. Of note, because the enums have inner structs it was difficult to not duplicate the error message.

Resolve: #42

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: CC0-1.0
 
+use core::fmt;
+
 /// Formats error.
 ///
 /// If `std` feature is OFF appends error source (delimited by `: `). We do this because
@@ -20,4 +22,73 @@ macro_rules! write_err {
             }
         }
     }
+}
+
+/// Hex decoding error.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum HexToBytesError {
+    /// Non-hexadecimal character.
+    InvalidChar(u8),
+    /// Purported hex string had odd length.
+    OddLengthString(usize),
+}
+
+impl fmt::Display for HexToBytesError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use HexToBytesError::*;
+
+        match *self {
+            InvalidChar(ch) => write!(f, "invalid hex character {}", ch),
+            OddLengthString(ell) => write!(f, "odd hex string length {}", ell),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for HexToBytesError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use HexToBytesError::*;
+
+        match self {
+            InvalidChar(_) | OddLengthString(_) => None,
+        }
+    }
+}
+
+/// Hex decoding error.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum HexToArrayError {
+    /// Conversion error while parsing hex string.
+    Conversion(HexToBytesError),
+    /// Tried to parse fixed-length hash from a string with the wrong length (got, want).
+    InvalidLength(usize, usize),
+}
+
+impl fmt::Display for HexToArrayError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use HexToArrayError::*;
+
+        match *self {
+            Conversion(ref e) => crate::write_err!(f, "conversion error"; e),
+            InvalidLength(got, want) =>
+                write!(f, "bad hex string length {} (expected {})", got, want),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for HexToArrayError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use HexToArrayError::*;
+
+        match *self {
+            Conversion(ref e) => Some(e),
+            InvalidLength(_, _) => None,
+        }
+    }
+}
+
+impl From<HexToBytesError> for HexToArrayError {
+    #[inline]
+    fn from(e: HexToBytesError) -> Self { Self::Conversion(e) }
 }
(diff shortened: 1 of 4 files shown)


tcharding:
Man, I forget if struct errors are supposed to have `#[non_exhaustive]`?

src/iter.rs
```diff
@@ -93,8 +94,12 @@ impl<'a> io::Read for HexToBytesIter<'a> {
 
 /// `hi` and `lo` are bytes representing hex characters.
 fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, HexToBytesError> {
-    let hih = (hi as char).to_digit(16).ok_or(HexToBytesError::InvalidChar(hi))?;
-    let loh = (lo as char).to_digit(16).ok_or(HexToBytesError::InvalidChar(lo))?;
+    let hih = (hi as char)
+        .to_digit(16)
+        .ok_or(HexToBytesError::InvalidChar(InvalidCharError { invalid: hi }))?;
```
Kixunil, line 99: You can keep it short by writing `ok_or(InvalidCharError { invalid: hi })?;` thanks to the conversion made by `?`.

src/parse.rs
```diff
@@ -9,7 +9,7 @@ use crate::alloc::vec::Vec;
 use crate::iter::HexToBytesIter;
 
 #[rustfmt::skip]                // Keep public re-exports separate.
-pub use crate::error::{HexToBytesError, HexToArrayError};
+pub use crate::error::{HexToBytesError, HexToArrayError, InvalidLengthError, OddLengthStringError, InvalidCharError};
```
Kixunil, line 12: I'm not really sure we should export the innermost errors. It'll be rare that people will access them by name (they will call method on them instead).

src/parse.rs
```diff
@@ -58,7 +58,10 @@ macro_rules! impl_fromhex_array {
                     }
                     Ok(ret)
                 } else {
-                    Err(HexToArrayError::InvalidLength(2 * $len, 2 * iter.len()))
+                    Err(HexToArrayError::InvalidLength(InvalidLengthError {
+                        expected: 2 * $len,
+                        got: 2 * iter.len(),
+                    }))
```
Kixunil, line 64: Possibly simplify it by omitting `HexToArrayError::InvalidLength` and calling `into` instead. (this is probably a matter of taste, I personally use `.into()` but don't care much)

src/error.rs
```diff
@@ -49,29 +52,77 @@ impl std::error::Error for HexToBytesError {
...
+}
+
+impl From<OddLengthStringError> for HexToBytesError {
+    #[inline]
+    fn from(e: OddLengthStringError) -> Self { Self::OddLengthString(e) }
+}
+
+/// Invalid hex character.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
```
Kixunil, line 74: This is not needed when at least one field is *not*  `pub` but I could understand if someone wants to avoid us forgetting to add it in the future.

src/error.rs
```diff
@@ -49,29 +52,77 @@ impl std::error::Error for HexToBytesError {
...
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for InvalidCharError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
+}
+
+/// Purported hex string had odd length.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
```
Kixunil, line 92: Also not needed

Kixunil (approved):
ACK a05458f51753846937ce14389477ea52fff6f392

Just a few things that would be nice to clean up.

src/iter.rs
tcharding: I back and forth'ed on this a bunch of times (literally added and remove code). It works sometimes and in some places it didn't so I went with uniform and wrote it out fully everywhere. I"ll have another go though.

src/parse.rs
tcharding: I had that exact same thought this morning when I woke up :)

src/error.rs
tcharding: I'm inclined to put it on every struct regardless because it makes scanning the error types quicker. However, this code is plain old sloppy, I've been not adding `Copy` when using `non_exhaustive`.

src/iter.rs
tcharding: Oh I misspoke, you suggestion is clearly better. I was referring to the `into` case below.

tcharding:
All suggestions implemented @Kixunil except removing `non_exhaustive`, ok with that?

src/error.rs
Kixunil: I think naming error type `*Error` is much better. There are other types that need to be `non_exhaustive` and are not error types.

Kixunil (approved):
ACK af04208bdba3add23e0722c94348832061820818

src/error.rs
tcharding: I don't follow your comment, all errors are suffixed with `Error`. I'll remove the `non_exhaustive` from all struct errors that exist solely to hide internals. On further thought today it is strange to have unneeded `non_exhaustive` and there will always be private fields since that is the whole point of these types.

apoelstra (approved):
ACK af04208bdba3add23e0722c94348832061820818

tcharding:
Please don't merge yet, I want to remove the unnecessary `non_exhaustive`.

apoelstra:
lol @tcharding you were just a bit too late, sorry

tcharding:
Ha! I thought you might have started the merge process soon as you acked. I'll fix as a follow up.

src/error.rs
Kixunil: Yes, they are and should continue to be. That's my point `rg 'struct [A-Za-z0-9]*Error' is better than `rg non_exhaustive`

PR #45, opened 2023-11-07
Remove `non_exhaustive` from error structs

tcharding:
Currently we have a few public error structs the sole purpose of which is to hide error internals. As such they have private (`pub(crate)`) fields and cannot be constructed by lib users, they do not need the `non_exhaustive` attribute for this reason.


tcharding:
This is follow up to #44

apoelstra (approved):
ACK da79c219e0a8ade5d7ff8f90cd184d3364709b39

PR #52, opened 2023-12-05
Revamp `BufEncoder`

Kixunil:
Due to MSRV limitation we had to previously supply the buffer externally. This can be now avoided since const generics are available. This removes the possibility of having the buffer allocated separtely from encoder but this is practically never needed. All uses construct both of them on stack, use them to encode something and then drop them.

To further improve the code, this uses `ArrayString` from the `arrayvec` crate (privately). This simplifies the implementation and provides better performance since that crate uses `unsafe` to handle uninitialized bytes and doesn't need double UTF-8 check (the one we have should get optimized-out).

Unfortunately, it's still not possible to implement the `AsHex` trait for all arrays but this at least simplifies the `DisplayArray` type. To avoid accidental panics its `new` method was made private and it's now constructed only with `as_hex` method. This also improves forward compatibility since we could one day change `new` to accept array instead (when const generics can accept expressions) without breaking (most of) the code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ std = ["alloc"]
 alloc = []
 
 [dependencies]
+arrayvec = { version = "0.7", default-features = false }
 core2 = { version = "0.3.2", default-features = false, optional = true }
 serde = { version = "1.0", default-features = false, optional = true }
 
--- a/examples/wrap_array_display_hex_trait.rs
+++ b/examples/wrap_array_display_hex_trait.rs
@@ -65,7 +65,7 @@ impl FromHex for Wrap {
 //     fn hex_reserve_suggestion(self) -> usize { self.0.as_ref().hex_reserve_suggestion() }
 // }
 impl<'a> DisplayHex for &'a Wrap {
-    type Display = DisplayArray<core::slice::Iter<'a, u8>, [u8; 64]>;
-    fn as_hex(self) -> Self::Display { DisplayArray::new(self.0.iter()) }
+    type Display = DisplayArray<'a, 64>;
+    fn as_hex(self) -> Self::Display { self.0.as_hex() }
     fn hex_reserve_suggestion(self) -> usize { 64 }
 }
--- a/src/buf_encoder.rs
+++ b/src/buf_encoder.rs
@@ -10,188 +10,25 @@
 
 use core::borrow::Borrow;
 
+use arrayvec::ArrayString;
+
 use super::Case;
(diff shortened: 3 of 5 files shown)


Kixunil:
Oh, we need to bump MSRV first.

tcharding (approved):
ACK f67817e29799bbe5304a4f606042f7cf0bbb8af5

tcharding:
Nice PR.

Kixunil:
I realized there's one more crazy optimization possible: take `NonNull<u8>` in `DisplayArray::new` (making it `unsafe` fn) and then just computing the length instead of storing it. But since I expect most things to inline it probably doesn't help much. (And it's tiny.)

apoelstra:
Looks like we need a rebase. I'm happy to accept this with or without your extra optimization (though if it forces users to use unsafe code, maybe I don't like it).

tcharding:
This PR broke `hashes::sha256`. I can't seem to work out how to display backwards now. Can you give me a tip pleaes?

``rust
                let v self.0.iter().rev().collect();
            pub fn backward_hex(&self) -> impl '_ + core::fmt::LowerHex + core::fmt::UpperHex {
                $crate::hex::display::DisplayArray::<_, [u8; $bits / 8 * 2]>::new(self.0.iter().rev())
            }
```

Now one cannot create a `DisplayArray` type.

apoelstra:
Hmm, gross. We may need a `display_hex_backward` which calls an `Encoder::put_bytes_backward`.

tcharding:
Came back to post https://github.com/rust-bitcoin/rust-bitcoin/pull/2473 and saw your suggestion, I'll have play with it.

Kixunil:
> This PR broke `hashes::sha256`

How?! Hashes use the `fmt_hex_exact` macro which was updated in this commit and does the right thing.

tcharding:
Oh sweeeet, `fmt_hex_exact` accepts an iterator still. Damn, wish I saw that yesterday.

Kixunil:
Ah, I see now. I think that rather than blowing up the API with a bunch of arbitrary methods we should just have `DisplayByteIter` with guts almost equal to those of `DisplayByteSlice` (and `DisplayByteSlice being just type alias setting `core::slice::Iter<'a, u8>` parameter) and allow constructing it from any iterator. It will slightly worsen the perf of backwards displaying but I guess it might be net-positive given we use `ArrayString` now. The alternative is to make some of the APIs worse.

Issue #50, opened 2023-12-05
Are we happy with `FromHex` usage of  `Err`?


tcharding:
Our `FromHex` trait has a trait method `from_hex` which has an associated error type `Err`. We used `Err` instead of `Error`, IIRC, because `from_hex` is close to `from_str` and `FromStr` uses `Err`.

I opened up the file today and was surprised to see `Err` not `Error` - thought I'd just check that, a few months on, we still think this was the correct decision? Not something we want to change after a 1.0 release.

Kixunil:
I prefer `Error`. One specific advantage is editors will correctly highlight it.

apoelstra:
I could go either way. Looks like `serde` uses `Error` so maybe we ought to copy them.

tcharding:
One bloke surprised, one with a preference, and one on the fence, that sounds like rough consensus to me, `Error` it is :)

PR #54, opened 2023-12-06
Change `FromHex` associated error type to `Err`

tcharding:
Currently we have `Err` as an associated error type for the `FromHex` trait. `Error` is [arguably] cleaner and definitely not worse.

This is the first API changing PR since we introduced the public API  files. 

Patch 1 makes the change, patch 2 updates the API files.

Resolve: #50

### DIFF
--- a/api/all-features.txt
+++ b/api/all-features.txt
@@ -203,7 +203,7 @@ pub enum hex_conservative::parse::HexToArrayError
 pub enum hex_conservative::parse::HexToBytesError
 pub fn &'a alloc::vec::Vec<u8>::as_hex(self) -> Self::Display
 pub fn &'a alloc::vec::Vec<u8>::hex_reserve_suggestion(self) -> usize
-pub fn alloc::vec::Vec<u8>::from_byte_iter<I>(iter: I) -> core::result::Result<Self, Self::Err> where I: core::iter::traits::iterator::Iterator<Item = core::result::Result<u8, hex_conservative::HexToBytesError>> + core::iter::traits::exact_size::ExactSizeIterator + core::iter::traits::double_ended::DoubleEndedIterator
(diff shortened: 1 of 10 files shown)


api/all-features.txt
```diff
@@ -298,8 +298,8 @@ pub fn hex_conservative::DisplayHex::hex_reserve_suggestion(self) -> usize
 pub fn hex_conservative::DisplayHex::to_hex_string(self, case: hex_conservative::Case) -> alloc::string::String
 pub fn hex_conservative::DisplayHex::to_lower_hex_string(self) -> alloc::string::String
 pub fn hex_conservative::DisplayHex::to_upper_hex_string(self) -> alloc::string::String
-pub fn hex_conservative::FromHex::from_byte_iter<I>(iter: I) -> core::result::Result<Self, Self::Err> where I: core::iter::traits::iterator::Iterator<Item = core::result::Result<u8, hex_conservative::HexToBytesError>> + core::iter::traits::exact_size::ExactSizeIterator + core::iter::traits::double_ended::DoubleEndedIterator
-pub fn hex_conservative::FromHex::from_hex(s: &str) -> core::result::Result<Self, Self::Err>
+pub fn hex_conservative::FromHex::from_byte_iter<I>(iter: I) -> core::result::Result<Self, Self::Error> where I: core::iter::traits::iterator::Iterator<Item = core::result::Result<u8, hex_conservative::HexToBytesError>> + core::iter::traits::exact_size::ExactSizeIterator + core::iter::traits::double_ended::DoubleEndedIterator
+pub fn hex_conservative::FromHex::from_hex(s: &str) -> core::result::Result<Self, Self::Error>
```
Kixunil, line 302: Oh, no, `cargo public-api` duplicates all trait items all over the place. That's stupid.

Kixunil (approved):
ACK b59f2e3081d2b28f74e790377a404739f114e83a

api/all-features.txt
apoelstra: Hmm, yeah. I wonder if we should file a bug/feature request with them, or just try to implement a post-processing step. Brutal to see a 250+ line diff just from changing one widely-implemented trait.

apoelstra: OTOH, it does "correctly" reveal that changing a single trait effectively changes the API of a gazillion types. So I dunno.

Kixunil: Yes, keep the types but not methods - those are redundant.

PR #55, opened 2023-12-06
Bump MSRV to Rust 1.56.1

tcharding:
As we are doing in other crates bump the MSRV to Rust v1.56.1


tcharding:
No sweat, I usually have time in the afternoon for simple changes that don't require too much brain power. Happy to help.

PR #56, opened 2023-12-11
Depend on our new io crate

tcharding:
This shows we can trivially depend on the new `bitcoin-io` crate soon as it is released. Draft til then.

Remove the conditional dependency on `core2` and unconditionally depend on the new `bitcoin-io` crate.


Kixunil (changes requested):
NACK, seeing `bitcoin-io` in the dep tree of a supposedly general-purpose crate is weird and will waste time of bunch of people who will then search why it's pulling in an unrelated thing.

I could accept an optional feature which will be always activated from `bitcoin` and other crates that depend on this. I think we can have feature gates on `bitcoin-io` | `std` and things should work.

Further I propose to **not** rename `bitcoin-io` to `io` in this crate to avoid confusion of non-bitcoin people.

Kixunil:
> bitcoin-io doesn't actually have anything to do with bitcoin, its just an io crate that is more maintained than core2.

Then it should've been called something else, unrelated to Bitcoin. I know it has nothing to do with Bitcoin but whoever stumbles across it will not know that. I for instance often check out crates that I suspect are in my tree for no reason and if I was in this situation I'd jump from my chair upon seeing this. And now take into account there are people "allergic to crypto"... (I know, I know...)

> or `io` should not be in the repo and should not be called `bitcoin-io`

This one. But if we eventually switch to `lgio` this resolves itself. :)

> what goes in them is 100% determined by the `rust-bitcoin` usecase.

I'm pretty sure we've added APIs that are not used in `bitcoin`. E.g. some iterators. Which we did for completeness.


Anyway, to resolve this, is there any specific problem with making it optional? We can later switch it to something else in a breaking release but if we're going to release a general-purpose crate let's not force-clutter it with `bitcoin-` stuff if not required.

tcharding:
> Anyway, to resolve this, is there any specific problem with making it optional? 

If that is all it takes lets go! So just to clarify:

1. your fine with a `bitcoin-io` dependency iff its optional? 
2. We can forget the "bitcoin in dep tree" bit? 
3. Then when `lgio` comes in that goes away too?

(numbered list so you can "yes, yes, no" or whatever)

Kixunil:
1. Yes, it'll get my ACK once also the tests pass etc
2. I don't understand this - are you pointing out that the dependency is still there? While an optional dependency is in a dep tree you won't see `Compiling bitcoin-io` when compiling a non-bitcoin not-io-using project and IIRC you won't see it in `cargo tree either` - that's my main issue with it. Seeing `bitcoin-io` in a bitcoin project is obviously fine.
3. Yes

tcharding:
`DO_FEATURE_MATRIX=true contrib/test.sh` runs successfully locally.

tcharding:
Re-done, to use the new release of `bitcoin-io`.

Cargo.toml
```diff
@@ -28,6 +28,7 @@ alloc = []
 arrayvec = { version = "0.7", default-features = false }
 
 core2 = { version = "0.3.2", default-features = false, optional = true }
+io = { package = "bitcoin-io", version = "0.1.1", default-features = false, optional = true }
```
Kixunil, line 31: Please don't rename this.

tcharding: Sure I'll remove it, whats the reason?

tcharding:
I'm not going to bother re-spinning this until #2499 goes in. Hitting this problem again after hitting it yesterday is making me want to not use a computer.

tcharding:
I just realized we need `BufRead` as well.

tcharding:
Changes in force push:
- Removed re-name of `bitcoin-io`
- Added docs to link the respective readers

Cargo.toml
Kixunil: I've already explained it somewhere else - someone who's not from `bitcoin` would find the code confusing.

apoelstra:
Looks like we need a rebase again.

tcharding:
If we are going to implement an `IterReader` in `rust-bitcoin` then we do not need this PR. Draft till I play further with https://github.com/rust-bitcoin/rust-bitcoin/pull/2039

Kixunil (changes requested):
Damn, my review didn't get posted again :angry: 

I'd rather not add it if we don't need it.

tcharding:
I'm confident we don't need this. The thought that we did comes from before we had `BufRead`.

src/iter.rs
```diff
@@ -41,6 +36,20 @@ impl<'a> HexToBytesIter<'a> {
...
+
+    fn read(&mut self, buf: &mut [u8]) -> usize {
+        let mut bytes_read = 0usize;
+        for dst in buf {
+            match self.next() {
+                Some(Ok(src)) => {
+                    *dst = src;
+                    bytes_read += 1;
+                }
+                _ => break,
```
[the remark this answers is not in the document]
apoelstra: You can take a [`Peekable<I>`](https://doc.rust-lang.org/std/iter/struct.Peekable.html) which sorta covers the usecases of a `PeekableIterator` trait.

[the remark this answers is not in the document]
Kixunil: Yes, but passing `Peekable<core::slice::Iter<'_, T>>` is just silly.

PR #58, opened 2024-01-08
Add support for prefixes to the `FromHex` trait

tcharding:
Draft because error handling of `FromHex` needs attention and also I can't work out how to print backwards in `hashes` anymore.

Add support for parsing prefixed strings as well as requiring the presence or absence of the prefix.

This code was copied from `rust-bitcoin`. The error type was re-named.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -5,9 +5,10 @@
 use core::{fmt, str};
 
 #[cfg(all(feature = "alloc", not(feature = "std")))]
-use crate::alloc::vec::Vec;
+use crate::alloc::{string::String, vec::Vec};
 use crate::error::InvalidLengthError;
 use crate::iter::HexToBytesIter;
+use crate::write_err;
 
 #[rustfmt::skip]                // Keep public re-exports separate.
 pub use crate::error::{HexToBytesError, HexToArrayError};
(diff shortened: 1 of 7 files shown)


apoelstra:
concept ACK. I'd drop the bolding from "not" in "not recommended for user input".

tcharding:
Added a solution to the inner `String` error, WIP for feedback, probably falls in @Kixunil's wheelhouse.

Kixunil (commented):
I was just writing an objection but I've just realized the `from_hex` method was wrong from the beginning of this crate. It changed the semantics of to allow `0x` which previously wasn't allowed causing everyone who used it to accidentally implement the broken "robustness" principle. While we've strictly issued breaking versions it's still subtle and people probably thought it's the same method.

Thankfully it didn't affect most of the hashes because of reversed handling. But it does affect `FromStr for Midstate`.

And since the robustness principle is broken, I would prefer changing the semantics of `from_hex` to what it originally was and use `from_maybe_prefixed_hex` for the current one.

Definitely concept ACK doing something about this.

Kixunil:
Regarding `InputString`, it was never intended to appear in stable public API and pulling in `bitcoin_internals` sucks for the same reason this crate is not called `bitcoin_hex`. So I suggest just creating `#[non_exhaustive] pub struct MissingPrefixError(#[cfg(feature = "alloc")] String);` and using that one in the variant.

tcharding:
> The main difference being that hex-encoded bytes never start with 0x prefix so trying to write that code would be silly/buggy.

Quote for #2323

I had definitely forgotten this, this is a simple discreet problem that we can fix. Leave it with me.

src/parse.rs
```diff
@@ -22,9 +23,64 @@ pub trait FromHex: Sized {
...
+    ///
+    /// Accepts an input string either with `0x` prefix or without, if you require specific handling
+    /// of the prefix see [`Self::from_prefixed_hex`] and [`Self::from_no_prefix_hex`].
+    #[rustfmt::skip]
+    fn from_maybe_prefixed_hex<
+        #[cfg(feature = "alloc")] S: AsRef<str> + Into<String>,
+        #[cfg(not(feature = "alloc"))] S: AsRef<str>,
+    >(s: S) -> Result<Self, Self::Error> {
+        if s.as_ref().starts_with("0x") {
+            Self::from_no_prefix_hex(s.as_ref().trim_start_matches("0x"))
```
Kixunil, line 48: `trim_start_matches` removes repeated matches so it allows `0x0x0x0x42`, which is wrong. Also the whole code seems to be needlessly complicated. You should only need this:

```rust
Self::from_no_prefix_hex(s.strip_prefix("0x").unwrap_or(s))
```

src/parse.rs
```diff
@@ -22,9 +23,64 @@ pub trait FromHex: Sized {
...
         I: Iterator<Item = Result<u8, HexToBytesError>> + ExactSizeIterator + DoubleEndedIterator;
 
+    /// Parses provided string as hex explicitly requiring there to not be a `0x` prefix.
+    ///
+    /// This is not recommended for user-supplied inputs because of possible confusion with decimals.
+    /// It should be only used for existing protocols which always encode values as hex without 0x prefix.
+    #[rustfmt::skip]
+    fn from_hex<
+        #[cfg(feature = "alloc")] S: AsRef<str> + Into<String>,
+        #[cfg(not(feature = "alloc"))] S: AsRef<str>,
```
Kixunil, line 33: This is broken because people can call the method for non-alloc and then alloc would silently add the bound.

src/parse.rs
```diff
@@ -22,9 +23,64 @@ pub trait FromHex: Sized {
...
+            Self::from_no_prefix_hex(s.as_ref().trim_start_matches("0x"))
+        } else {
+            Self::from_no_prefix_hex(s)
+        }
+    }
+
+    /// Parses provided string as hex explicitly requiring there to not be a `0x` prefix.
+    ///
+    /// This is not recommended for user-supplied inputs because of possible confusion with decimals.
+    /// It should be only used for existing protocols which always encode values as hex without 0x prefix.
```
Kixunil, line 57: There should be a note that this is equivalent to `from_hex` But maybe let's not duplicate the method?

src/parse.rs
```diff
@@ -22,9 +23,64 @@ pub trait FromHex: Sized {
...
+    >(s: S) -> Result<Self, PrefixedError<Self::Error>> {
+        use PrefixedError::*;
+
+        if !s.as_ref().starts_with("0x") {
+            #[cfg(feature = "alloc")]
+            return Err(MissingPrefix(MissingPrefixError(s.into())));
+            #[cfg(not(feature = "alloc"))]
+            return Err(MissingPrefix(MissingPrefixError()));
+        } else {
+            Ok(Self::from_no_prefix_hex(s.as_ref().trim_start_matches("0x"))?)
```
Kixunil, line 82: Same problem as above, you can use `s.as_ref().strip_prefix("0x").ok_or_else(|| MissingPrefixError(s.into()).into())`

Kixunil (changes requested):

tcharding:
This PR is tied up with the `FromHex` error stuff.

tcharding:
Consumed by #67

PR #59, opened 2024-01-25
Improve the check public API script

tcharding:
Most importantly - actually run it in CI. Face palm.

Also add missing call to `uniq` and function-ize the script.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -102,3 +102,18 @@ jobs:
         run: cargo install cross --locked
       - name: run cross test
         run: cross test --target s390x-unknown-linux-gnu
+
+  API:
+    name: Check for changes to the public API
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v3
+      - name: Checkout Toolchain
+        uses: dtolnay/rust-toolchain@nightly
+      - name: Install cargo-public-api
+        run: cargo install --locked cargo-public-api
+      - name: Running API checker script
+        run: ./contrib/check-for-api-changes.sh
--- a/api/all-features.txt
+++ b/api/all-features.txt
@@ -1,71 +1,12 @@
-impl<A, B> core::marker::Send for hex_conservative::display::DisplayArray<A, B> where A: core::marker::Send, B: core::marker::Send
-impl<A, B> core::marker::Sync for hex_conservative::display::DisplayArray<A, B> where A: core::marker::Sync, B: core::marker::Sync
-impl<A, B> core::marker::Unpin for hex_conservative::display::DisplayArray<A, B> where A: core::marker::Unpin, B: core::marker::Unpin
-impl<A, B> core::panic::unwind_safe::RefUnwindSafe for hex_conservative::display::DisplayArray<A, B> where A: core::panic::unwind_safe::RefUnwindSafe, B: core::panic::unwind_safe::RefUnwindSafe
-impl<A, B> core::panic::unwind_safe::UnwindSafe for hex_conservative::display::DisplayArray<A, B> where A: core::panic::unwind_safe::UnwindSafe, B: core::panic::unwind_safe::UnwindSafe
-impl<A: core::clone::Clone + core::iter::traits::collect::IntoIterator, B: hex_conservative::buf_encoder::FixedLenBuf> core::fmt::Debug for hex_conservative::display::DisplayArray<A, B> where <A as core::iter::traits::collect::IntoIterator>::Item: core::borrow::Borrow<u8>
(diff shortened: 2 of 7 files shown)


apoelstra:
Can't reproduce this test failure locally, even after upgrading to the latest nightly :/.

tcharding:
That is exactly my confusion. Locally this PR is correct but when running on the CI VM we are getting a different sort order.

tcharding:
This now mimics what I did in `bech32` https://github.com/rust-bitcoin/rust-bech32/pull/141

I"ve squashed the whole thing down into a single commit, because I"m hacking on the weekend and need to stop now. Can re split it up on Monday if required.

apoelstra:
I think the squashing is fine; I use `git diff --stat` and then diff individual files for these so that I can skim the API files but carefully read the others. It's a little annoying but not a big deal.

apoelstra:
But what doesn't seem fine is that it appears you've emptied all the .txt files :). I think you can't use `--unique` with other options and you need to manually pipe into `uniq`.

.github/workflows/rust.yaml
```diff
@@ -102,3 +102,18 @@ jobs:
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v3
+      - name: Checkout Toolchain
+        uses: dtolnay/rust-toolchain@nightly
+      - name: Install cargo-public-api
+        run: cargo install --locked cargo-public-api
```
Kixunil, line 117: Would be nice to cache this.

tcharding:
>  I think you can't use --unique with other options and you need to manually pipe into uniq.

Ha! And while doing this PR I was like "hmm seems `--unique` is working now" - no clue how I came to that conclusion, epic fail.

tcharding:
I've split the PR up and been a bit more careful. The big patch (last one) is now refactor only.

tcharding:
Ok, now we are back to the same problem we had days ago. The sort order produced when running the script on the CI VM is different to the sort order produced when running it locally on my machine - go figure?

I have checked the version of `bash` and or `sort` on the VM and they both match my machine.

apoelstra:
Maybe the locale differs? From my `sort` manpage

```
       *** WARNING *** The locale specified by the environment affects sort order.  Set LC_ALL=C to get the traditional sort order
       that uses native byte values.
```

tcharding:
Hot tip! I'll investigate, thanks.

tcharding:
Oh my goodness, it works.

apoelstra:
Lol. Everything may be a file, a text file even, but after 55 years of UNIX development I guess just there wasn't enough demand for some canonical sort order. :roll_eyes: 

Can you squash this stuff into a nicer PR :)

tcharding:
> Can you squash this stuff into a nicer PR :)

Just move the changes in the last patch to a better place, right?

tcharding:
Another lesson should probably be learned - RTFM ... all the way to the bottom.

apoelstra:
> Another lesson should probably be learned - RTFM ... all the way to the bottom.

I'm curious when and why that warning was added to my manpage. I'll bet it is a fairly recent addition and I'll bet it doesn't show up in many `sort` manpages.

apoelstra:
>Just move the changes in the last patch to a better place, right?

Could you squash it so that we introduce the final script in one commit, run it in another (or in the first, that's fine too), and add CI in another?

Right now pretty-much every commit is "fix the script somehow and re-run it".

tcharding:
Done (even though that makes reviewing the api file changes impossible IMO).

apoelstra:
> Done (even though that makes reviewing the api file changes impossible IMO).

Because this is an initial run, I had no intention of reviewing the API file changes beyond "does it look like a lot of API surface is there and can I reproduce it".

tcharding:
No sweat, lets get it in.

apoelstra:
Heh, locally I get

```
  ┊24┊+impl core::marker::StructuralEq for hex_conservative::Case
  ┊25┊+impl core::marker::StructuralEq for hex_conservative::HexToArrayError
  ┊26┊+impl core::marker::StructuralEq for hex_conservative::HexToBytesError
```

which looks like it's a new nightly thing? I'm surprised CI didn't complain. My nightly is from 2024-01-20.

apoelstra:
"Structural equality" BTW is the property, roughly, where two types are equal if you can destructure (match) and eventually land on the same enum variants. It's useful for formal proofs because structural equality is something a prover can reason about, unlike the "user implemented a function called `eq` and it returned `true`" notion of equality.

apoelstra:
Oh, ran with a nightly from tonight and now it runs clean. Ok, sure.

apoelstra (approved):
ACK 4148fd7f50d0db6ed3d3a7023c897665e3d23354

Kixunil (approved):
ACK 4148fd7f50d0db6ed3d3a7023c897665e3d23354

PR #62, opened 2024-02-11
Return specific error

tcharding:
The `HexToByesIter::new` constructor can only fail in a single manner, return the specific error type for this failure instead of the more general error enum.

### DIFF
--- a/api/all-features.txt
+++ b/api/all-features.txt
@@ -1,41 +1,54 @@
 impl core::clone::Clone for hex_conservative::Case
 impl core::clone::Clone for hex_conservative::HexToArrayError
 impl core::clone::Clone for hex_conservative::HexToBytesError
+impl core::clone::Clone for hex_conservative::OddLengthStringError
 impl core::cmp::Eq for hex_conservative::Case
 impl core::cmp::Eq for hex_conservative::HexToArrayError
 impl core::cmp::Eq for hex_conservative::HexToBytesError
+impl core::cmp::Eq for hex_conservative::OddLengthStringError
 impl core::cmp::PartialEq for hex_conservative::Case
 impl core::cmp::PartialEq for hex_conservative::HexToArrayError
 impl core::cmp::PartialEq for hex_conservative::HexToBytesError
+impl core::cmp::PartialEq for hex_conservative::OddLengthStringError
 impl core::convert::From<hex_conservative::HexToBytesError> for hex_conservative::HexToArrayError
+impl core::convert::From<hex_conservative::OddLengthStringError> for hex_conservative::HexToBytesError
 impl core::default::Default for hex_conservative::Case
 impl core::error::Error for hex_conservative::HexToArrayError
 impl core::error::Error for hex_conservative::HexToBytesError
+impl core::error::Error for hex_conservative::OddLengthStringError
 impl core::fmt::Debug for hex_conservative::Case
 impl core::fmt::Debug for hex_conservative::HexToArrayError
 impl core::fmt::Debug for hex_conservative::HexToBytesError
+impl core::fmt::Debug for hex_conservative::OddLengthStringError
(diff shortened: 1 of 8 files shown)


apoelstra:
Nice, first test of the API script :).

It is a bit noisy but I think on net it's helpful. It's good to see everything that we're committing to in the API in one place.

apoelstra (approved):
ACK 090e007f84fe98f263411bc1d0577951f0ccd551

api/all-features.txt
```diff
@@ -1,41 +1,54 @@
 impl core::clone::Clone for hex_conservative::Case
 impl core::clone::Clone for hex_conservative::HexToArrayError
 impl core::clone::Clone for hex_conservative::HexToBytesError
+impl core::clone::Clone for hex_conservative::OddLengthStringError
```
Kixunil, line 4: What the hell? The previous commit didn't add `Clone`.

Kixunil (commented):
ACK the code change, the API change looks broken.

api/all-features.txt
apoelstra: Innnteresting.

So, previously `OddLengthStringError` was not actually exposed except as the inner type of a variant of `HexToBytesError`. Rust did not complain about this because it is a `pub struct`, it was just in a non-`pub` module. IMO this was a bug in our old API, and maybe one we could programmatically check for now that we have the API txt files.

This patch properly exports `OddLengthStringError`, and with it all the existing trait impls on it.

This is an old old issue with Rust's module system, but since it's abused to seal traits etc it's a WONTFIX.

Kixunil: Oh, damn! I wish there was a first-class support for sealing so we wouldn't need to deal with this crap. Or at least some attributes for these tools to distinguish the cases. (It's not entirely WONTFIX because they could just add first-class sealing and turn those cases into deny-by-default lints. But it won't happen anytime soon.)

tcharding: This threw me as well, I believe all these additions are because the type is now available at the new location `hex_conservative::`.

tcharding:
> It is a bit noisy but I think on net it's helpful.

Perhaps we try to do re-exports as separate PRs because they are always going to be noisy? Or separate patches in a PR with multiple updates to the API files?

tcharding:
I found it helpful, interestingly I've found myself reaching for it in other crates that don't have it yet.

src/parse.rs
```diff
@@ -24,7 +24,8 @@ pub trait FromHex: Sized {
 
     /// Produces an object from a hex string.
     fn from_hex(s: &str) -> Result<Self, Self::Error> {
-        Self::from_byte_iter(HexToBytesIter::new(s)?)
+        let iter = HexToBytesIter::new(s).map_err(Into::into)?;
```
Kixunil, line 27: When digging why this is not ambiguous I found that the whole API is likely broken around error handling. Will do more digging tomorrow.

api/all-features.txt
Kixunil: Might be worth calling it out in the commit/PR description.

src/parse.rs
tcharding: I tend to agree. I don't have an exact reason but all my recent attempts other than this one to improve the error handling of the `FromHex` trait have failed. Looking forward to seeing what you come up with.

api/all-features.txt
tcharding: My bad, should have done that originally. Now included in the patch for the api changes.

tcharding:
FTR both @Kixunil and I think the `FromHex` trait error handling is borked but this patch is objectively correct and is a step forward.

apoelstra (approved):
ACK addf681d78fff48385bd1d2b9d4fbe67ea4ebfb4

tcharding:
Lets just roll this into #67.

PR #63, opened 2024-02-11
Re-export all errors from crate root

tcharding:
We don't expect users to use the modules when importing types from this crate, therefore all public errors should be re-exported.


tcharding:
I forgot, we don't re-export the inner error types.

Issue #61, opened 2024-01-27
Impl `FromHex` for all array lengths and use `ArrayVec` (or `[MaybeUninit<u8>; $len]`)


Kixunil:
MSRV 1.56 allows these impls, so we should use them. Further, internally we should use `ArrayVec` or `MaybeUninit`. I prefer `arrayvec::ArrayVec` for now (note it's not the one from `bitcoin-internals` mainly to avoid `bitcoin-` pollution).

PR #66, opened 2024-02-12
Implement FromHex for all array lengths

tcharding:
Now we have const generics in our MSRV we can implement `FromHex` for all array lengths using `arrayvec::ArrayVec` which we allready depend on.

This code is not optimised, there is almost certainly a better way but it my brain is being recalcitrant at the moment.

Fix: #61


Kixunil (approved):
ACK 45078e229d4c72d70fffebc1c4fbb9dd8af89691

I see nothing unoptimized here. Maybe it's possible to just call `collect` but I highly doubt it can improve perf in this code. Just has less LOC.

tcharding:
Nice, thanks.

apoelstra (approved):
ACK 45078e229d4c72d70fffebc1c4fbb9dd8af89691 love to see that red in the API files

PR #67, opened 2024-02-13
Fix the error handling in the `FromHex` trait

tcharding:
On top of #62.

BOOM, got something done.


src/parse.rs
```diff
@@ -8,36 +8,65 @@ use arrayvec::ArrayVec;
...
 
-    /// Produces an object from a hex string.
-    fn from_hex(s: &str) -> Result<Self, Self::Error> {
-        Self::from_byte_iter(HexToBytesIter::new(s)?)
+    /// Produces an object from a hex string that may or may not include a `0x` prefix.
+    ///
+    /// Equivalent to [`Self::from_maybe_prefixed_hex`].
+    fn from_hex(s: &str) -> Result<Self, FromHexError<Self::Error>> {
+        Self::from_maybe_prefixed_hex(s)
+    }
```
tcharding, line 32: Flagging this, this is not the functionality you talked about @Kixunil, feels to me like `from_hex` should accept 0x or not. Also, since its ambiguous I feel there is some merit to having bath short and long forms.

Kixunil: The reason I wanted to go back is that most likely 99% of people who updated to our new crate got their code silently broken and they don't know about it (it wasn't in the changelog because nobody noticed). So I think we should restore it. I could also accept removing this ambiguous method entirely which forces explicit handling.

src/error.rs
```diff
@@ -28,51 +30,139 @@ macro_rules! write_err {
...
     #[inline]
-    fn from(e: InvalidCharError) -> Self { Self::InvalidChar(e) }
+    fn from(e: OddLengthStringError) -> Self { Self::OddLengthString(e) }
+}
+
+/// Error decoding a hex string that explicitly excludes a prefix.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FromNoPrefixHexError<E> {
+    /// Input string contains a prefix.
+    Prefix(ContainsPrefixError),
```
Kixunil, line 73: In principle this shouldn't be needed. A `0x` prefix is just invalid digit `x`. It is true that handling the special case can be helpful in error messages but we have the information in the input string, no need to store it (and in case of no-alloc we can store a `bool`).

src/parse.rs
tcharding: Or maybe that isn't strong enough. Alternately we could remove `from_hex` for a couple of releases then consider re-introducing it - same aim but more aggressive.

src/error.rs
```diff
@@ -28,51 +30,139 @@ macro_rules! write_err {
 
 /// Hex decoding error.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum HexToBytesError {
-    /// Non-hexadecimal character.
-    InvalidChar(InvalidCharError),
+pub enum FromHexError<E> {
```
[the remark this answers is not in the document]
tcharding: This is an alternate approach to how we currently have it, using the `From` trait bound on `Self::Error`.

`examples::custom` has an example of how I thought this would be used:
```rust
impl FromHex for ALittleBitHexy {
    type Error = CustomError;

    fn from_byte_iter<I>(iter: I) -> Result<Self, Self::Error>
    where
        I: Iterator<Item = Result<u8, InvalidCharError>> + ExactSizeIterator + DoubleEndedIterator,
    {
        // Errors if the iterator is the wrong length.
        let data = <[u8; 32] as FromHex>::from_byte_iter(iter).map_err(CustomError::Hex)?;

        // An example of some application specific error.
        if data == [0; 32] {
            return Err(CustomError::AllZeros);
        }

        // This is a contrived example (using x==0).
        Ok(ALittleBitHexy { data, x: 0 })
    }
}
```

> Shouldn't it be the other way around? Have InvalidChar variant and generic InvalidLength variant where the InvalidLength would be either OddLengthError or InvalidLengthError.

I didn't fully process this bit. I can have a play with this idea if the custom example above doesn't change you view.

src/error.rs
tcharding: Agreed but, at least in my experience, getting the invalid character error is a bit jarring, contains prefix error is more informative. Since this is a hopefully write-once-change-very-little crate I thought the thoroughness was worth it.

tcharding:
Added the input string to  errors an removed the todos.

tcharding:
Open question, are you guys ok with multiple commits with the same git brief message `Update API files`, both in this PR and in the git index in general. Given that the commits mechanical changes to the API text files only, and will never need to be part of a debugging git bisect session.

src/error.rs
tcharding: I've pulled all the prefix logic out of this PR. Reducing the scope to just the error stuff.

src/parse.rs
Kixunil: I just wouldn't put it back.

src/error.rs
Kixunil: As I said, you don't need a separate variant for it.

apoelstra:
> Open question, are you guys ok with multiple commits with the same git brief message `Update API files`, both in this PR and in the git index in general. Given that the commits mechanical changes to the API text files only, and will never need to be part of a debugging git bisect session.

Fine by me, yep.

apoelstra:
In 79e853411a75268059eeaa7bf2688540aac09b34 the example files' implementation of `std::error::Error` should return the hex error from the `source` method. But I won't hold up the PR over this.

apoelstra (approved):
ACK c5aeb7d7f1a942df0abd30aefd1dbf9ce5367ad2

src/error.rs
tcharding: Also marking as resolved because of the relevance to prefixes.

tcharding:
> But I won't hold up the PR over this

Since these examples are designed so users can cut'n'past them I'd prefer to have it right.

tcharding:
@Kixunil what do you rekon of this PR now? It fixes the issue we had with hex to array error having invalid length and odd length but the array `from_hex` is a bit clunky still returning two errors the same. And its not as nice now that one has to implement both traits where as before, as you originally designed it, one only had to implement `from_byte_iter`, I believe that it is better now though so that errors are handled at the correct place. Your thoughts?

src/parse.rs
```diff
@@ -2,54 +2,78 @@
...
 
 /// Trait for objects that can be deserialized from hex strings.
 pub trait FromHex: Sized {
-    /// Error type returned while parsing hex string.
-    type Error: From<HexToBytesError> + Sized + fmt::Debug + fmt::Display;
+    /// Error type returned while constructing type from byte iterator.
+    type FromByteIterError: fmt::Debug + fmt::Display;
+
+    /// Error type returned by `from_hex`.
+    type FromHexError: From<Self::FromByteIterError> + fmt::Debug + fmt::Display;
```
Kixunil, line 24: Having two errors is quite complicated and I'm not convinced that it achieves anything. In most cases people are bubbling an error up the call stack so one error is better and in those other cases just matching on enum is fine.

I think we should either have:

```rust
enum ParseError<E = Infallible> {
    InvalidChar(InvalidCharError),
    InvalidLength(InvalidLengthError),
    InvalidValue(E),
}
```

Or have `type Error: From<InvalidCharError>`.

I think what we really need though is to decouple invalid char from length error so that invalid length can vary. And I was thinking of having an API that doesn't return `InvalidLength` for guaranteed-valid lengths.

Kixunil (commented):
Looks needlessly complicated to me and I'm not even sure what it tries to achieve. I might look into error decoupling myself if you have hard time figuring it out.

tcharding:
I'm happy for you to take a look and see what you come up with. It is high priority that we get _something_ working because at the moment moving any more types to `units` is blocked on `hex`, FTR there are a few problems and I don't know exactly what the minimum amount of changes to get `hex` released is. I could easily of bitten off more than is necessary.

Kixunil:
One more thought: are prefixes in array-like hex things actually used anywhere? I've never seen a transaction, hash or other byte array entity written as `0x...`. So maybe it makes sense to just get rid of all that prefix stuff entirely?

apoelstra:
@Kixunil they are easy to imagine using for locktimes.

But no, no "array-like" structures or anything where the hex is encoding a bytestring.

Kixunil:
> they are easy to imagine using for locktimes.

Yes, I mean do it for those types but not for array-like ones.

tcharding:
Closing in favour of #75

PR #69, opened 2024-02-13
Add `FromHex::as_hex_backwards`

tcharding:
Done on top of #67 and #71

Just the second to last patch (last is API changes), draft for feedback please.

Adds functionality to support printing slices and vectors of bytes backwards as used by `rust-bitcoin`. The current code is definitely a bit klunky.


Kixunil:
I don't understand the motivation, just format the reversed iterator.

tcharding:
Closing, turns out we can just handle backwards in `rust-bitcoin`.

Issue #70, opened 2024-02-13
`max` is divided by 2


tcharding:
EDIT: Probably not worth reading this issue, just see #71

I'm pretty sure this line of code is wrong but I can't come up with a unit test to prove it. Patching it to what I think it should be does not break any unit tests.

https://github.com/rust-bitcoin/hex-conservative/blob/7116a460ad2dbf6b2aea9f37ebc933db79927c07/src/display.rs#L169

We are dividing the max string length by 2 as well as multiplying the number of bytes by 2 - we should just do one (I believe).

```rust
            let string_len = match f.precision() {
                Some(max) if self.bytes.len() * 2 > (max + 1) / 2 => max,
                Some(_) | None => self.bytes.len() * 2,
            };
```

Should be I think

```rust
```rust
            let string_len = match f.precision() {
                Some(max) if self.bytes.len() * 2 > max => max,
                Some(_) | None => self.bytes.len() * 2,
            };
```

tcharding:
ooo, `(max + 1) / 2` might have been to handle odd length max width, which makes no sense for hex strings. So perhaps this should be
```rust
            let string_len = match f.precision() {
                // Do ceil division on max so as to ignore odd length
                // max width which makes no sense for hex strings.
                Some(max) if self.bytes.len() > (max + 1) / 2 => max,
                Some(_) | None => self.bytes.len() * 2,
            };
```

tcharding:
Sorry for the noise, took me a while to work out the interaction between `width` and `precision` - TIL.

PR #71, opened 2024-02-13
Truncate using even length hex string

tcharding:
There are a couple of problems with the width/precision code when we truncate:

- We are printing one too many hex characters if truncation is odd
- The nested precision code (inside width) is doing the wrong comparison

Includes unit tests to verify the behaviour of the new code.

Fix: #70

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -154,7 +154,6 @@ pub struct DisplayByteSlice<'a> {
 
 impl<'a> DisplayByteSlice<'a> {
     fn display(&self, f: &mut fmt::Formatter, case: Case) -> fmt::Result {
-        use fmt::Write;
         // There are at least two optimizations left:
         //
         // * Reusing the buffer (encoder) which may decrease the number of virtual calls
@@ -166,8 +165,14 @@ impl<'a> DisplayByteSlice<'a> {
 
         let pad_right = if let Some(width) = f.width() {
             let string_len = match f.precision() {
-                Some(max) if self.bytes.len() * 2 > (max + 1) / 2 => max,
-                Some(_) | None => self.bytes.len() * 2,
+                #[rustfmt::skip]
+                Some(max) => {
+                    // Odd max width which makes no sense for hex strings, honour max but use even length.
+                    let max = if max % 2 == 1 { max - 1 } else { max };
+                    let hex_length = self.bytes.len() * 2;
+                    if hex_length > max { max } else { hex_length }
+                }
+                None => self.bytes.len() * 2,
             };
 
             if string_len < width {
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 2205dc7c0bbc157180ba1f4fc228e4407ef7b2e5

src/display.rs
```diff
@@ -166,8 +165,14 @@ impl<'a> DisplayByteSlice<'a> {
 
         let pad_right = if let Some(width) = f.width() {
             let string_len = match f.precision() {
-                Some(max) if self.bytes.len() * 2 > (max + 1) / 2 => max,
-                Some(_) | None => self.bytes.len() * 2,
+                #[rustfmt::skip]
+                Some(max) => {
+                    // Odd max width which makes no sense for hex strings, honour max but use even length.
```
tcharding, line 170: Is this statement correct? Yesterday I realised that we may want to support parsing integers in which case `0xa` should be parsable, precision and width should allow `format!("{0:.1}"` in that case, right?

apoelstra: Ah, yeah, I think if this is supposed to parse integers (which seems to be the case, given that we're supporting the `0x` prefix) then we ought not to require even lengths.

tcharding: Oh boy, the last two hours has show me that parsing hex integers is a whole can of worms. I am starting to see why it was never included in this crate.

src/display.rs
```diff
@@ -166,8 +165,14 @@ impl<'a> DisplayByteSlice<'a> {
...
+                    let max = if max % 2 == 1 { max - 1 } else { max };
```
Kixunil, line 171: Branchless: `let max = max - (max % 2);` (Maybe the compiler does it I'm not sure.)

Kixunil (commented):
Why should we truncate more than what was specified? IMO truncating part of hex digit is valid.

src/display.rs
tcharding: That suggestion is hot. Legendary.

tcharding:
> Why should we truncate more than what was specified? IMO truncating part of hex digit is valid.

Without doing more invasive changes we cannot support displaying an odd number of hex characters. Therefore it seems more correct to me to print one less than max instead of one more than max, since max is AFAIU an upper limit on how many characters should be displayed.

tcharding:
Used the branchless suggestion and updated the code comment.

apoelstra (approved):
ACK c7b006902d2c2127fe87327dcd611dcf5281f67f though I agree it would be better if we could support odd-length strings

tcharding:
Changes in force push: `s/ceil/floor/` in code comment, woops.

Also I checked that this does not conflict with #75.

apoelstra (approved):
ACK 0d7b18ca21da510b1e944427a51ef596082fa221 though I agree it would be better if we could support odd-length strings

tcharding:
Will need rebasing after #77 goes in.

tcharding:
I don't believe this conflicts with #75 but if @Kixunil acks you could merge his first just in case - this one is non-urgent.

tcharding:
Rebased, just for good measure.

tcharding:
Converting to draft because I don't care, I just want a release.

PR #73, opened 2024-02-15
Add support for parsing integer types

tcharding:
Draft for feedback please, and also because on top of #67.

Add support to the library for parsing standard integer types u8, u16, u32, and u64.
    
Note we allow but do not enforce 0x prefix.

If this gets a concept ack we could then do `from_maybe_prefixed_hex, `from_prefixed_hex`, and `from_no_prefix_hex` for explicit handling if wanted.


tcharding:
There is no appetite for this at the moment.

Issue #60, opened 2024-01-27
`HexToArrayError` should not wrap `HexToBytesError`


Kixunil:
`HexToBytesError` contains odd length variant which is unreachable when parsing array - it's just invalid length.

We should put invalid char directly into `HexToArrayError` and panic during conversion if necessary.

tcharding:
I agree that the issue is correct but posit that the current code is still technically correct.  If fixing as suggested makes the generics any more complicated than they already are then I think it should be left as is.

Kixunil:
I don't see how it has anything to do with generics. Just modify the variants of the enum and panic during conversion. Unless you want to avoid the panic but we can still do that without complicating the *public* API by calling into an inner function after checking the length which only returns errors for invalid characters.

tcharding:
This issue is conceptually correct but not technically correct, currently an odd length string returns the `HexToBytesError::OddLengthString` variant. As can be seen by this existing test:
```rust
        assert_eq!(
            <[u8; 4]>::from_hex(oddlen),
            Err(HexToBytesError::OddLengthString(OddLengthStringError { len: 17 }).into())
        );
```

tcharding:
As I suspected this issue is curly. I've had a bunch of goes at it. I am sick right now though.

tankyleo:
@tcharding I think we can close this issue - it was fixed in #80

tcharding:
Mad, thanks.

PR #75, opened 2024-02-18
Error cleanup

Kixunil:
Potential alternative to error improvements in #67, based on his work.
The commit message from the most significant commit:

The `ArrayFromHexError` contained `FromBytesError` which had a redundant
`OddLengthStringError` since it implies `InvalidLength`. Decoupling them
was quite difficult due to interaction with the `FromHex` trait. However
the `FromHex` trait had another weirdness: it required implementing
`from_byte_iter` which took already-parsed bytes. That looked fine in
theory however it lead to complications and the only benefit of it was
sparing *one* line of code.

In the ideal world we'd have `FromIterator<T> for Result<[T; N], _>` and
then `FromHex` would naturally be implemented for all things that can be
constructed from an iterator of bytes. However such impl is far away and
we prefer supporting arrays today since they are very common.

Thus this change deletes `from_byte_iter` method and makes `from_hex`
mandatory. It also removes useless bouds on associated type `Error` and
finally decouples the error types.

Notably, this changes error returned from the `FromHex` impl on arrays
to return just `InvalidLength` even if the lenght is odd. The logic
first checks the length so the compiler hopefully sees that the length
is always even and can elide the check in `HexToBytesIter::new` after
inlining.

Fix: #60

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -10,10 +10,10 @@ use std::io;
 #[cfg(all(feature = "core2", not(feature = "std")))]
 use core2::io;
 
-use crate::error::{InvalidCharError, OddLengthStringError};
+use crate::error::InvalidCharError;
 
 #[rustfmt::skip]                // Keep public re-exports separate.
-pub use crate::error::HexToBytesError;
+pub use crate::error::{HexToBytesError, OddLengthStringError};
 
 /// Iterator over a hex-encoded string slice which decodes hex and yields bytes.
 pub struct HexToBytesIter<'a> {
@@ -34,9 +34,9 @@ impl<'a> HexToBytesIter<'a> {
     ///
     /// If the input string is of odd length.
     #[inline]
-    pub fn new(s: &'a str) -> Result<HexToBytesIter<'a>, HexToBytesError> {
+    pub fn new(s: &'a str) -> Result<HexToBytesIter<'a>, OddLengthStringError> {
         if s.len() % 2 != 0 {
-            Err(OddLengthStringError { len: s.len() }.into())
+            Err(OddLengthStringError { len: s.len() })
         } else {
             Ok(HexToBytesIter { iter: s.bytes() })
         }
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -11,10 +11,7 @@ use std::io;
 #[cfg(all(feature = "core2", not(feature = "std")))]
 use core2::io;
 
-use crate::error::InvalidCharError;
-
-#[rustfmt::skip]                // Keep public re-exports separate.
-pub use crate::error::OddLengthStringError;
+use crate::error::{InvalidCharError, OddLengthStringError};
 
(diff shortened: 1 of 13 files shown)


Kixunil:
FTR the last commit is kinda off-topic. I thought it can help removing errors when the input has known length but it's not *that* simple, so I gave up for now.

src/iter.rs
```diff
@@ -58,26 +63,34 @@ impl<'a> Iterator for HexToBytesIter<'a> {
         let (min, max) = self.iter.size_hint();
         (min / 2, max.map(|x| x / 2))
```
tcharding, line 64: This line needs updating mate to remove the div by 2.
```rust

    #[test]
    fn iter_length() {
        let hex = "deadbeef";
        let iter = HexToBytesIter::new(hex).unwrap();
        let (min, max) = iter.size_hint();
        assert_eq!(min, 4);
        assert_eq!(max, Some(4));
    }
```

src/iter.rs
```diff
@@ -10,46 +11,50 @@ use std::io;
 #[cfg(all(feature = "core2", not(feature = "std")))]
 use core2::io;
 
-use crate::error::{InvalidCharError, OddLengthStringError};
+use crate::error::InvalidCharError;
 
 #[rustfmt::skip]                // Keep public re-exports separate.
-pub use crate::error::HexToBytesError;
+pub use crate::error::{HexToBytesError, OddLengthStringError};
```
tcharding, line 17: clippy warns about `HexToBytesError` and I think we should re-export `InvalidCharError` because it is public in `HexToBytesIter` (returned by `next`).

tcharding (commented):
This is awesome, thanks for sorting it out!

tcharding:
>  I wrote this PR somewhat quickly to get the main idea out because I wasn't sure if my solution is the best.

I'm  travelling home today, when I get there I plan on re-doing the "test next hex release in rust-bitcoin" PR so we can see how this looks.

tcharding:
Will need #77 and also formatting.

tcharding:
https://github.com/rust-bitcoin/rust-bitcoin/pull/2473 is done man and IMO the changes in this PR are good.

Kixunil:
I really don't want to bother regenerating API changes - see #78

tcharding:
@apoelstra please merge #75 so @Kixunil can rebase this one, this PR is holding up the whole show - we can't do anything in `units` until `hex` is released.

tcharding (approved):
~ACK a169c96fda7713e2963c1218c76cc0bd3874b7f5~

I rebased this and removed the API changes and checked it all locally.

tcharding (approved):
ACK c94bf01d3c64160b77f8486520cae13b509e4635

tcharding:
Just needs rebase and remove 61baef84c1b1c82e7027c091fe35a00f9a209271

PR #64, opened 2024-02-11
Release tracking PR: `v0.2.0`

tcharding:
Add changelog entry and bump the version.

Note that during the 0.1.1 release I updated `fuzz/generate-files.sh` incorrectly so it has 0.2.0 in it already.

See commit: `42f3cda Bump version to 0.1.1`

### Testing

Tested in a rust-bitcoin POC PR: https://github.com/rust-bitcoin/rust-bitcoin/pull/2473.

PR #76, opened 2024-02-21
Make things public

tcharding:
In order to get the current code to build as a dependency of `rust-bitcoin` we need a few more things to be public. The hider error types and also the `iter::HexDigitIter`.

If we do not want to pollute the crate level API with the less used things, which all these are, then we need to make the `iter` and `error` modules public so niche users can get access to them.

In regards to the hider errors, we either have to make the fields public or add getters for them because we want the inner values in `rust-bitcoin` - here we elect to make the fields public and add `non_exhaustive` to maintain forward compatibility.


Kixunil:
I believe I already did resolve this in my latest push. Is anything missing? `HexDigitIter` is internal, so I only reexported type alias and I believe it'd be nicer to have the iterators in `parse` and `encode` modules rather than iter. (Just like e.g. `serde` has `ser` and `de` only.)

Kixunil:
I'd prefer accessors in errors too.

tcharding:
Addedd accessors and made fields `pub(crate)` (removing necessity for `non_exhaustive`).

tcharding:
> I believe I already did resolve this in my latest push. Is anything missing? 

Oh yes, this had gotten a bit stale. Now only adds accessor functions and makes `error` pubilc.

src/lib.rs
```diff
@@ -46,7 +46,7 @@ extern crate alloc;
 
 pub mod buf_encoder;
 pub mod display;
-mod error;
+pub mod error;
```
Kixunil, line 49: Should this be `parse::error`? There are no non-parsing errors.

src/error.rs
```diff
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: CC0-1.0
```
Kixunil, line 1: The commit message still refers to iterators.

src/error.rs
```diff
@@ -92,6 +99,11 @@ pub struct OddLengthStringError {
     pub(crate) len: usize,
 }
 
+impl OddLengthStringError {
+    /// Returns the odd length of the input string.
+    pub fn odd_length(&self) -> usize { self.len }
```
Kixunil, line 104: I think it should be just `length`.

src/lib.rs
tcharding: `InvalidCharError` and `OddLengthStringError` are `iter` things.  We could group parse/iter separately to display/buf_encoder but for this little crate I don't think its worth it.

tcharding:
Note to reviewers, the `InvalidCharError` and `OddLengthStringError` are not hider errors, even though they appear as such in `HexToBytesError` - the `InvalidLengthError` is the only hider error in this crate.

src/error.rs
tcharding: Thanks, fixed.

src/error.rs
tcharding: Done, thanks.

src/lib.rs
Kixunil: Those are still parsing error but for some reason in the `iter` module. I think we should split the modules based on intended use not based on technical details.

tcharding: Bro, is it really necessary to hold up the release for this discussion - you are taking the piss right now.

tcharding:
Justification for merge:
- PR has an ACK
- We do not have a two-ack policy in this crate
- Only outstanding concern is about module structure which is not at all important for a 0.2.0 release
- We are waiting on the `hex` release for a bunch of work in `rust-bitcoin`
- I want to rebase the release tracking PR #64 on top of this so we don't have to do another 24 merge/rebase/push cycle.
- Its version 0.2.0 for fucks sake.

src/lib.rs
Kixunil: No, we can publish them later. I don't understand why we should do it right now.

PR #79, opened 2024-02-22
Remove API infrastructure

tcharding:
There is push back against the recently merged API stuff. Its a burden and if not everyone is onboard we can't really do it.

Lets just get work done and merge some PRs already.

Fix: #75


apoelstra:
Fine, if that's what we need to have a release.

PR #86, opened 2024-05-10
Fix arrayvec minimal version.

jeandudey:
The ArrayString<CAP>::remaining_capacity method was introduced in arrayvec@0.7.2:

https://github.com/bluss/arrayvec/commit/481f93084eecebda0445cb7f81b40b9789dd1559

This causes `cargo build -Z minimal-versions' to fail when building hex-conservative.

* Cargo.toml (dependencies) <arrayvec>: Update to 0.7.2.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ std = ["alloc"]
 alloc = []
 
 [dependencies]
-arrayvec = { version = "0.7", default-features = false }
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK f8d0e854b8665d4b236a47f6189ff6354c1dc590 thanks, good catch!

jeandudey:
Thanks!

PR #88, opened 2024-05-12
Fix order of 'got' and 'want' fields when populating InvalidLength error

### LINKED ISSUE #87: In v0.1.1, InvalidLength error prints 'expected' and  'got' values swapped
While during parsing the values are in order (want, got):

https://github.com/rust-bitcoin/hex-conservative/blob/bd9c2c7b44bc89e91ec6bb4f7164672e27a43f64/src/parse.rs#L88

`Display` prints them in order (got, want):

https://github.com/rust-bitcoin/hex-conservative/blob/bd9c2c7b44bc89e91ec6bb4f7164672e27a43f64/src/parse.rs#L124-L134

This was, probably inadvertently, later [fixed](https://github.com/rust-bitcoin/hex-conservative/pull/44/files#diff-5583e398a8deec11b274d9965eff8b5ade5226c7a020ca214a27d7d07dcc8a29L89-R62) and v0.2.0 does not suffer from it anymore.

The issue is that rust-bitcoin v0.31.2 depends on hex-conservative v0.1.1 and therefore the error messages in v0.31.2 are incorrect. I know it's too tiny an issue, but if there is will to have this fixed (I can help), perhaps the solution would be to publish v0.1.2 with fix and let rust-bitcoin v0.31.3 depend on it?

---

Background: I am upgrading rust-lightning to rust-bitcoin v0.31 (see [#2745](https://github.com/lightningdevkit/rust-lightning/issues/2745)) and it so happens that rust-lightning has a [test](https://github.com/lightningdevkit/rust-lightning/blob/d3cd0802532d869fdc2b1a620cc76b647895cc54/lightning-block-sync/src/convert.rs#L604-L626) which fails with v0.31 due to this issue. I have not yet opened PR on rust-lightning, may do so tomorrow.

```
failures:

---- convert::tests::into_txid_from_json_response_with_invalid_hex_data stdout ----
thread 'convert::tests::into_txid_from_json_response_with_invalid_hex_data' panicked at lightning-block-sync/src/convert.rs:610:17:
assertion `left == right` failed
  left: "bad hex string length 64 (expected 6)"
 right: "bad hex string length 6 (expected 64)"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- convert::tests::into_txid_from_json_response_with_invalid_txid_data stdout ----
thread 'convert::tests::into_txid_from_json_response_with_invalid_txid_data' panicked at lightning-block-sync/src/convert.rs:622:17:
assertion `left == right` failed
  left: "bad hex string length 64 (expected 4)"
 right: "bad hex string length 4 (expected 64)"


failures:
    convert::tests::into_txid_from_json_response_with_invalid_hex_data
    convert::tests::into_txid_from_json_response_with_invalid_txid_data
```

jirijakes:
Error `HexToArrayError::InvalidLength` has two numeric fields 'got' and 'want' (in this order), however the two fields were populated in the opposite order. This patch fixes it and thus fixes #87.

To crosscheck the order of fields:

https://github.com/rust-bitcoin/hex-conservative/blob/bd9c2c7b44bc89e91ec6bb4f7164672e27a43f64/src/parse.rs#L115-L122

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -85,7 +85,9 @@ macro_rules! impl_fromhex_array {
                     }
                     Ok(ret)
                 } else {
-                    Err(HexToArrayError::InvalidLength(2 * $len, 2 * iter.len()))
+                    let got = 2 * iter.len();
+                    let want = 2 * $len;
+                    Err(HexToArrayError::InvalidLength(got, want))
                 }
             }
         }
@@ -183,7 +185,9 @@ mod tests {
     fn hex_to_array_error() {
         use HexToArrayError::*;
         let len_sixteen = "0123456789abcdef";
-        assert_eq!(<[u8; 4]>::from_hex(len_sixteen), Err(InvalidLength(8, 16)));
+        let result = <[u8; 4]>::from_hex(len_sixteen);
+        assert_eq!(result, Err(InvalidLength(16, 8)));
+        assert_eq!(&result.unwrap_err().to_string(), "bad hex string length 16 (expected 8)");
     }
 
     #[test]


tcharding (approved):
ACK bba0a6c88734b1e96086763bc2ee7a32c5109aac

tcharding:
The two failing CI jobs are ok and unrelated to this PR, its because the target branch is old. @apoelstra will also verify my claims when he runs his local CI before merging. Thanks for the PR @jirijakes.

FTR this PR has a few more lines in the diff than I would have done myself but it does the job 100% and is only going to exist on the `0.1.x` release branch since `master` has moved on already.

apoelstra (approved):
ACK bba0a6c88734b1e96086763bc2ee7a32c5109aac

apoelstra:
Looks good! Note that we need a separate PR to bump the version number if we want to release this.

tcharding:
Done in https://github.com/rust-bitcoin/hex-conservative/pull/89

PR #90, opened 2024-05-16
Add `impl_fmt_traits` macro

tcharding:
Add a macro that implements the `fmt::{LowerHex, UpperHex, Display, Debug}` traits. Includes support for displaying backwards.

### Context

In `bitcoin_hashes` we have _two_ versions of this code, moving it here is better because it is  general hex related code. Note however that the reason for supporting display backwards might see strange unusual to the casual reader.

## Note

Note please the use of naked `core::foo` in the macro requiring `core` to be in scope, is this reasonable of should we do the ugly re-export stuff we do in `bitcoin_hashes`: `$crate::_export::_core::foo`?

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -343,6 +343,168 @@ macro_rules! fmt_hex_exact {
 }
 pub use fmt_hex_exact;
 
+/// Adds `core::fmt` trait implementations to type `$ty`.
+///
+/// Implements:
+///
+/// - `fmt::{LowerHex, UpperHex}` using [`fmt_hex_exact`].
+/// - `fmt::{Display, Debug}` by calling `LowerHex`.
+///
+/// Requires:
+///
+/// - `$ty` must implement `IntoIterator<Item=Borrow<u8>>`.
+///
+/// ## Parameters
+///
+/// * `$ty` - the type to implement traits on.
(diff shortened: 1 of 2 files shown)


tcharding:
PR is tested in https://github.com/rust-bitcoin/rust-bitcoin/pull/2770

apoelstra:
In a06b7ed236a56b881f1b60f213fea04880bde4c4:

Concept ACK. But

* We probably should do the ugly `core` stuff, since it guards against the case where a user has their own `core` module in scope.
* We should change the interface to look more Rust-like, e.g.

```rust
impl_fmt_traits! {
    #[reverse(true)] // or whatever
    impl<T: Whatever> all_fmt_traits for MyType<T> {
        const LENGTH: usize = 32;
    }
}
```

tcharding:
woops, macro stuff sill  to come.

apoelstra (approved):
ACK cb9b86a13b7b16c5dc166b08131146bb68f2b75f I am a *tiny* bit tempted to say that the `const LENGTH` should wind up as an actual associated constant. Though this is easy to add in a followup PR

Kixunil:
@apoelstra that's impossible because you can't write `[u8; Self::LENGTH * 2]` and declaring the constant anyway is weird since none of the traits have it.

PR #92, opened 2024-06-20
Introduce more serde utils

tcharding:
Copy the `serde_utils` modules for ser/deser of `BTreeMap`s. Currently includes the `hex_bytes` module even though it would seem we don't need it but I couldn't get it to work with the current functions. Its private as opposed to the original where it was public.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,11 +27,10 @@ alloc = []
 [dependencies]
 arrayvec = { version = "0.7.2", default-features = false }
 core2 = { version = "0.3.2", default-features = false, optional = true }
-serde = { version = "1.0", default-features = false, optional = true }
+serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
 
 
 [dev-dependencies]
-serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 
 [[example]]
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -106,3 +106,304 @@ where
         d.deserialize_map(HexVisitor(PhantomData))
     }
 }
+
+pub(crate) struct SerializeBytesAsHex<'a>(pub(crate) &'a [u8]);
+
+impl<'a> serde::Serialize for SerializeBytesAsHex<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        use crate::DisplayHex;
+
+        serializer.collect_str(&format_args!("{:x}", self.0.as_hex()))
+    }
+}
+
+#[cfg(feature = "alloc")]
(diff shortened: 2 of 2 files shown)


tcharding:
If this lands we can use this to assist the crate smashing effort in `rust-bitcoin`. The new crates need these utils.

apoelstra:
To clarify -- you copied this from rust-bitcoin? Or from std?

apoelstra:
Ah, it's from rust-bitcoin. Cool.

apoelstra:
@tcharding can you update Cargo.toml to turn on the `derive` feature of `serde` even when it's not a dev-dependency?

I'm not sure why cargo seems to be OK with the current PR but it's technically incorrect and my local non-cargo tooling is breaking on it.

tcharding:
Added an additional patch up front.

Kixunil (changes requested):
~~I believe the first commit is just plain wrong~~ (see below) and second has questionable usefulness.

Side note: I never intended for `hex-conservative` to be a dumping ground for random hex-related shit. I want it to be a nice idiomatic crate. Adding stuff that's only used in one crate until someone else needs it is not helpful to achieve this goal.

Kixunil:
> my local non-cargo tooling is breaking on it.

Oh, this PR causes it, it wasn't a problem before. Having it in two commits is confusing.

So if this is accepted I'd prefer to squash the commits since they don't make sense separated. That being said, adding another dependency for this is annoying, it'd be nice trying to impl it manually since it just uses `serialize_with` anyway.

apoelstra:
> Side note: I never intended for `hex-conservative` to be a dumping ground for random hex-related shit

Should we start `bitcoin-hex` then with hex-related shit that we need for rust-bitcoin?

Kixunil:
This particular one is only needed for PSBT. If we need it for multiple things and multiple crates then that would be a strong indication that putting it into `hex-conservative` is actually worth it. (But we could let it sit for a bit in `internals`.)

apoelstra:
@tcharding is it only PSBT?

Actually, if it's really only PSBT then we can maybe get away with dropping the serde serialization entirely and just serializing as strings... I really dislike breaking serde de/serialization but this one is really ad-hoc.

Kixunil:
Exactly. Serialize as base64.

tcharding:
Used also in `witness`, I'll investigate if we can remove that.

bitcoin/src/blockdata/witness.rs:476:                seq.serialize_element(&crate::serde_utils::SerializeBytesAsHex(elem))?;

Kixunil:
Damn, I had a huge brain fart yesterday forgetting there's also `SerialzieBytesAsHex`. That one is generally useful I believe and should be added. Sorry about the noise.

tcharding:
For posterity, this was closed in favour of https://github.com/rust-bitcoin/hex-conservative/pull/96

PR #96, opened 2024-07-02
Enable serialization of byte slices

tcharding:
Add three types to facilitate serializing a byte slice as hex. The idea for this code is copied from `rust-bitcoin::serde_utils`.

Make the tersely named type use lowercase and add explicit lower and upper versions as we do for the serialize functions.

Tested in https://github.com/rust-bitcoin/rust-bitcoin/pull/2971

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -71,6 +71,42 @@ where
     }
 }
 
+/// Byte slice wrapper to serialize as a hex string in lowercase characters.
+pub struct SerializeBytesAsHex<'a>(pub &'a [u8]);
+
+impl<'a> serde::Serialize for SerializeBytesAsHex<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serialize_lower(self.0, serializer)
+    }
+}
+
+/// Byte slice wrapper to serialize as a hex string in lowercase characters.
+pub struct SerializeBytesAsHexLower<'a>(pub &'a [u8]);
+
+impl<'a> serde::Serialize for SerializeBytesAsHexLower<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serialize_lower(self.0, serializer)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


tcharding:
Based on discussion in https://github.com/rust-bitcoin/hex-conservative/pull/92

src/serde.rs
```diff
@@ -71,6 +71,42 @@ where
...
+
+impl<'a> serde::Serialize for SerializeBytesAsHexUpper<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serialize_upper(self.0, serializer)
+    }
+}
+
```
Kixunil, line 109: The indentation is missing.

Kixunil (commented):
Except for indentation it looks good.

src/serde.rs
tcharding: You sure? I can't see the problem neither can `rustfmt`?

Kixunil: LOL, it's an optical illusion caused by GH UI :D

src/serde.rs
```diff
@@ -71,6 +71,42 @@ where
     }
 }
 
+/// Byte slice wrapper to serialize as a hex string in lowercase characters.
+pub struct SerializeBytesAsHex<'a>(pub(crate) &'a [u8]);
```
Kixunil, line 75: Have you actually tested the API? I now see there's no way to create the struct. :)

I think we can simply make the field `pub`.

Kixunil (changes requested):
Needs constructors.

src/serde.rs
tcharding: Ha!

src/serde.rs
tcharding: Holy moly, what a crap PR. Sorry to waste your time.

tcharding: There are no invariants so I made the inner field `pub`.

Kixunil (approved):
ACK 87a2daf2097174ff318358b7d20cbbdff329b515

apoelstra (approved):
ACK 87a2daf2097174ff318358b7d20cbbdff329b515

PR #102, opened 2024-09-04
Bump MSRV to Rust 1.63.0

tcharding:
Rust 1.63.0 is in Debian stable, lets use it as our MSRV.

Includes:

- move of `msrv` field in `clippy.toml` to `rust-version` field in `Cargo.toml`
- required fuzz changes (toolchain, bump of `hongfuzz` version)
- gets fields by grepping when generating the fuzz manifest
- removes now unneeded pinning in CI script

And includes unrelated fix by adding `unexpected_cfgs`.


tcharding:
Needs #103

tcharding:
Woops, the date in the branch name is wrong, I did not do this PR in the future.

Kixunil (commented):
Concept ACK

tcharding:
Bloody hell, will try to fix tomorrow.

tcharding:
3 days to debug,  f**king `generate-files.sh` bit me again.

tcharding:
This fucking PR is starting to piss me off.

tcharding:
Compiler is version 1.58 even over in #106. Just kill me already :)

tcharding:
Phew, finally found it, all I needed was a breather and a good feed. FTR tt was buried in the `fuzz/Cargo.toml` output of `generate-files.sh`.

clippy.toml
```diff
@@ -1 +1 @@
-msrv = "1.56.1"
+msrv = "1.63.0"
```
Kixunil, line 1: This file should be no longer needed, `clippy` can read `Cargo.toml`

contrib/test.sh
```diff
@@ -3,7 +3,7 @@
 set -ex
 
 FEATURES="std alloc serde"
-MSRV="1\.48\.0"
+MSRV="1\.63\.0"
```
Kixunil, line 6: SSOT break. This should just read `Cargo.toml`.

fuzz/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "hex-fuzz"
-edition = "2018"
+edition = "2021"
+rust-version = "1.63.0"
```
Kixunil, line 4: I wish we could pull these from workspace already... It's just one version away.

fuzz/generate-files.sh
```diff
@@ -34,6 +35,12 @@ path = "$targetFile"
 EOF
 done
 
+cat >> "$REPO_DIR/fuzz/Cargo.toml" <<EOF
+
+[lints.rust]
+unexpected_cfgs = { level = "deny", check-cfg = ['cfg(fuzzing)'] }
+EOF
```
Kixunil, line 42: This is unrelated but not a big deal.

fuzz/generate-files.sh
```diff
@@ -11,7 +11,8 @@ source "$REPO_DIR/fuzz/fuzz-util.sh"
 cat > "$REPO_DIR/fuzz/Cargo.toml" <<EOF
 [package]
 name = "hex-fuzz"
-edition = "2018"
+edition = "2021"
+rust-version = "1.63.0"
```
Kixunil, line 15: Since this is generated we should pull the value from single source.

clippy.toml
tcharding: Cool, we can remove this all over the place if that is the case.

contrib/test.sh
tcharding: If #106 gets support this doesn't matter because the file is deleted, can we leave it for now?

fuzz/generate-files.sh
tcharding: Ah yes, you are correct.

fuzz/generate-files.sh
tcharding: Unrelated to this PR, right? Can we just put this in, I spent way too long on this extremely annoying PR already.

contrib/test.sh
Kixunil: OK, concept ACK doing #106

fuzz/generate-files.sh
Kixunil: It's literally just replacing the line with `$(grep '^rust-version ' "$REPO_DIR/Cargo.toml")`

tcharding:
Please see new PR description.

Kixunil:
I've modified the description to better describe what's going on.

.github/workflows/rust.yaml
```diff
@@ -54,15 +54,15 @@ jobs:
...
+    name: Test - 1.63.0 toolchain
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
     steps:
       - name: Checkout Crate
         uses: actions/checkout@v3
       - name: Checkout Toolchain
-        uses: dtolnay/rust-toolchain@1.56.1
+        uses: dtolnay/rust-toolchain@1.63.0
```
Kixunil, line 65: I'm not going to hold this because the PR is blocking a bunch of other work but this should be read from `Cargo.toml` in a previous step.

Kixunil (approved):
ACK 5cd37e394f7e859b48e64f0f14e6f7d1ccbe314d

tcharding:
Thanks, I did try to look up the `clippy` thing and couldn't find any docs, I did not want to claim something that I had not verified.

.github/workflows/rust.yaml
tcharding: Holla if you can think of a better way of tracking issues across "core crates": https://github.com/rust-bitcoin/rust-bitcoin/issues/3335

Kixunil: There are org-wide "projects" but I'm not sure if it's the right tool.

Kixunil:
I could find it easily: https://doc.rust-lang.org/nightly/clippy/lint_configuration.html#msrv

I just put `rust clippy configuration "rust-version"` to google to verify that clippy is reading the field and the reason I came up with the idea to check it directly is that it's absolutely sensible/expected for clippy to read it.

tcharding:
Lol, my search-foo needs improving.

tcharding:
Oh and its because we didn't have a `rust-version` until we started using edition 2021.

Kixunil:
It's not tied to `edition`, just to Rust version. But I don't remember which stabilized it and don't care to check.

apoelstra:
Updated my local CI so that it will use `rust-version` from Cargo.toml if it's present, and only if it's not, fall back to `msrv` in clippy.toml (and use 1.56.1 if that's missing).

apoelstra:
Hmm, this is gonna kill my performance for multi-workspace repos if I can't determine my compiler set until after I've split into multiple workspaces.

But ok, we can discuss that when we try to make that change for rust-bitcoin.

PR #103, opened 2024-09-04
fuzz: Add unexpected_cfgs to manifest

tcharding:
Quiet the new lint by adding an unexpected cfgs for `fuzzing`.

Issue #51, opened 2023-12-05
Depend on `bitcoin-io`


tcharding:
Once we release `bitcoin-io` we want to use it here, right? That would make `bitcoin-io` part of the public API which implies we need to do stable release of `bitcoin-io` before we can do a stable release of this crate.

Kixunil:
I think `bitcoin-io` will never be stable and there's no need to depend on it here. Yeah, `no_std` people don't get readers and writers but everyone else avoids a weird `bitcoin` dependency in a general-purpose crate. Our crates don't need readers and writers and if they do, we can add them to `bitcoin-io` instead.

Kixunil:
> Huh?

I hope `push_decode` will eventually obviate the need for it (before it stabilizes).

> I mean I dont have a need for them, but I assume others do?

Then who are those others? If they speak up, I'd like to work with them to figure out a good solution.

> I mean we can also just rename it to not say "bitcoin" on the tin.

Or rename `bitcoin-io` to `core3` :P

PR #104, opened 2024-09-04
Remove `core2` dependency

tcharding:
We use `core2` to implement `Read` because it needs the `io::Error`. Now that we have `bitcoin-io` we can implement `Read` from there and not provide the `core2` impl.

- Remove optional `core2` dependency.
- Add an optional dependency on `bitcoin-io`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,12 +21,12 @@ members = ["fuzz"]
 
 [features]
 default = ["std"]
-std = ["alloc"]
-alloc = []
+std = ["alloc", "bitcoin-io?/std"]
+alloc = ["bitcoin-io?/alloc"]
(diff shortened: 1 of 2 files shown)


tcharding:
Needs #102 and #103

tcharding:
I assume this PR is going to start the typical "there shouldn't be a bitcoin related crate in the dep tree"

tcharding:
Oh I did not see #51.

Kixunil:
NACK adding optional dependency on `bitcoin-io` because it just reintroduces the same problem core2 has.

apoelstra:
Given our very limited use of `io` I agree it'd be better to just drop it.

Issue #98, opened 2024-08-19
Store position of invalid char in `InvalidCharError`


Kixunil:
This can be very helpful when displaying to the user which char is invalid.

PR #107, opened 2024-09-08
Store position of invalid char in `InvalidCharError`

leo-nash:
Fix: #98

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -16,6 +16,7 @@ pub type HexSliceToBytesIter<'a> = HexToBytesIter<HexDigitsIter<'a>>;
 /// Iterator yielding bytes decoded from an iterator of pairs of hex digits.
 pub struct HexToBytesIter<T: Iterator<Item = [u8; 2]>> {
     iter: T,
+    original_len: usize,
 }
 
 impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
@@ -38,19 +39,26 @@ impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
     }
 }
 
-impl<T: Iterator<Item = [u8; 2]>> HexToBytesIter<T> {
+impl<T: Iterator<Item = [u8; 2]> + ExactSizeIterator> HexToBytesIter<T> {
     /// Constructs a custom hex decoding iterator from another iterator.
     #[inline]
-    pub fn from_pairs(iter: T) -> Self { Self { iter } }
+    pub fn from_pairs(iter: T) -> Self { Self { original_len: iter.len(), iter } }
 }
 
-impl<T: Iterator<Item = [u8; 2]>> Iterator for HexToBytesIter<T> {
+impl<T: Iterator<Item = [u8; 2]> + ExactSizeIterator> Iterator for HexToBytesIter<T> {
     type Item = Result<u8, InvalidCharError>;
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
         let [hi, lo] = self.iter.next()?;
-        Some(hex_chars_to_byte(hi, lo))
+        Some(hex_chars_to_byte(hi, lo).map_err(|(c, is_high)| InvalidCharError {
+            invalid: c,
+            pos: if is_high {
+                (self.original_len - self.iter.len() - 1) * 2
+            } else {
+                (self.original_len - self.iter.len() - 1) * 2 + 1
+            },
+        }))
     }
 
     #[inline]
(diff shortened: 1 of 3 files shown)


src/iter.rs
```diff
@@ -15,7 +15,7 @@ pub type HexSliceToBytesIter<'a> = HexToBytesIter<HexDigitsIter<'a>>;
 
 /// Iterator yielding bytes decoded from an iterator of pairs of hex digits.
 pub struct HexToBytesIter<T: Iterator<Item = [u8; 2]>> {
-    iter: T,
+    iter: Enumerate<T>,
```
Kixunil, line 18: I think it'd be better to store the original length of the string and then compute position as `original_len - iter.len() * 2`. The reason is we probably want a fast success path and having to increase a number every loop iteration in addition to increasing the pointer inside iterator is waste of performance.

This needs `ExactSizeIterator` bound but that should be fine.

Kixunil (commented):
Thanks for the PR! Even though I think it could be done differently, we might want to compare performance to test if my claim is correct.

src/iter.rs
```diff
@@ -62,30 +70,50 @@ impl<T: Iterator<Item = [u8; 2]>> Iterator for HexToBytesIter<T> {
     #[inline]
     fn nth(&mut self, n: usize) -> Option<Self::Item> {
         let [hi, lo] = self.iter.nth(n)?;
-        Some(hex_chars_to_byte(hi, lo))
+        Some(hex_chars_to_byte(hi, lo).map_err(|(c, h)| InvalidCharError {
```
tcharding, line 73: nit: Perhaps use `(c, is_high)`

src/iter.rs
```diff
@@ -153,9 +181,9 @@ impl<'a> ExactSizeIterator for HexDigitsIter<'a> {}
 impl<'a> core::iter::FusedIterator for HexDigitsIter<'a> {}
 
 /// `hi` and `lo` are bytes representing hex characters.
-fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, InvalidCharError> {
-    let hih = (hi as char).to_digit(16).ok_or(InvalidCharError { invalid: hi })?;
-    let loh = (lo as char).to_digit(16).ok_or(InvalidCharError { invalid: lo })?;
+fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, (u8, bool)> {
```
tcharding, line 184:
```suggestion
/// `hi` and `lo` are bytes representing hex characters.
///
/// Returns the valid byte or the invalid input byte and a bool indicating error for `hi` or `lo`.
fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, (u8, bool)> {
```

tcharding (approved):
Thanks for your contribution!

ACK 10571c434940951c481b39a77a8ff304e34bce52

My two suggestions are nits only, feel free to use or not.

tcharding (approved):
ACK 67556b3617ed7ff9b3e89f03e99a4e16a4f13db2

tcharding:
Nice one, thanks man.

leo-nash:
TIL `cargo clippy --tests`

tcharding:
`cargo clippy --all-targets` would get you there as well.

tcharding (approved):
ACK d7b4be55e2fd17eeb7be60ebfc6ad52ada88067f

apoelstra:
The `ExactSizeIterator` bound is actually pretty onerous. It means that users can no longer chain iterators, for example, since [`Chain` disables the trait for annoying reasons](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html).

I'm ok with this if Kix is (this isn't too hard to work around; typically you are only chaining iterators within encapsulated code where you can write your own non-crippled `Chain`). But want to draw attention to it.

apoelstra (approved):
ACK d7b4be55e2fd17eeb7be60ebfc6ad52ada88067f successfully ran local tests

Kixunil:
@apoelstra thanks for pointing that out! I've just realized the iterator is actually publicly constructible, I haven't noticed the public constructor before, so I was thinking it's not a big deal to change it.

We could roll our own poor-man's specialization as `serde_json` does. Do we want to?

apoelstra:
> We could roll our own poor-man's specialization as `serde_json` does. Do we want to?

You mean, having separate method names for exact-sized vs non-exact-sized iterators? Or something nicer than that?

Personally I think we shoud just accept this PR and accept that we have to do a major rev because of it (though we shouldn't do a release quickly until we've audited the rest of our errors at least).

Kixunil:
> having separate method names for exact-sized vs non-exact-sized iterators? Or something nicer than that?

Possibly both, but I mainly had `&str` optimization in mind. I've also realized that my originally suggested approach to handle `char` doesn't work so we probably need this specialization anyway for it.

> Personally I think we shoud just accept this PR and accept that we have to do a major rev because of it

Yes, I think that blocking the PR over this is too much. I was mainly asking because we might want to revert to the original design instead.

apoelstra:
> Possibly both, but I mainly had `&str` optimization in mind.

You mean, having separate methods for `&str` and for `&[u8]`?

>Yes, I think that blocking the PR over this is too much.

Ok, cool. I'll go ahead and merge it.

Kixunil:
> You mean, having separate methods for `&str` and for `&[u8]`?

Yes.

Kixunil (approved):
ACK d7b4be55e2fd17eeb7be60ebfc6ad52ada88067f

Issue #82, opened 2024-02-23
When displaying to limited precision we should internally stop encoding early


apoelstra:
In #81 we respect the `{:.10}` precision formatter, but do so by internally encoding a whole string then truncating before giving to the formatter. We should instead just stop the iterator early.

I don't believe this would be a breaking change in terms of semver, but note that this is externally visible since the user passes in their own iterator, and can tell how much of it we consume.

PR #108, opened 2024-09-10
Only encode the bytes formatted in the hex string

leo-nash:
Fix: #82

- This adds the `ExactSizeIterator` bound on the `bytes` parameter of `fmt_hex_exact_fn`. I assume this is ok because `fmt_hex_exact` calls `len` on `bytes` before calling `fmt_hex_exact_fn`, and `fmt_hex_exact_fn` is `#[doc(hidden)]` (so not officially part of the public API ?).
- With the `ExactSizeIterator` bound, I can wrap the `bytes` iterator in a `Take` and take the performance hit of incrementing a `usize` during encoding only when necessary.

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -510,6 +510,8 @@ pub use impl_fmt_traits;
 // Whether hex is an integer or a string is debatable, we cater a little bit to each.
 // - We support users adding `0x` prefix using "{:#}" (treating hex like an integer).
 // - We support limiting the output using precision "{:.10}" (treating hex like a string).
+//
+// This assumes `bytes.len() * 2 == N`.
 #[doc(hidden)]
 #[inline]
 pub fn fmt_hex_exact_fn<I, const N: usize>(
@@ -522,14 +524,17 @@ where
     I::Item: Borrow<u8>,
 {
     let mut encoder = BufEncoder::<N>::new();
-    encoder.put_bytes(bytes, case);
-    let encoded = encoder.as_str();
-
-    if let Some(precision) = f.precision() {
-        if encoded.len() > precision {
-            return f.pad_integral(true, "0x", &encoded[..precision]);
+    let encoded = match f.precision() {
+        Some(p) if p < N => {
+            let n = (p + 1) / 2;
+            encoder.put_bytes(bytes.into_iter().take(n), case);
+            &encoder.as_str()[..p]
         }
-    }
+        _ => {
+            encoder.put_bytes(bytes, case);
+            encoder.as_str()
+        }
+    };
     f.pad_integral(true, "0x", encoded)
 }
 
@@ -587,6 +592,8 @@ mod tests {
             let dummy = Dummy([42; 32]);
             assert_eq!(dummy.to_string(), "2a".repeat(32));
             assert_eq!(format!("{:.10}", dummy), "2a".repeat(5));
+            assert_eq!(format!("{:.11}", dummy), "2a".repeat(5) + "2");
+            assert_eq!(format!("{:.65}", dummy), "2a".repeat(32));
         }
 
         #[test]


apoelstra (approved):
ACK 27023e7475e7536febb42567e19880a8afea15f9 successfully ran local tests

src/display.rs
```diff
@@ -520,16 +522,21 @@ pub fn fmt_hex_exact_fn<I, const N: usize>(
 where
     I: IntoIterator,
     I::Item: Borrow<u8>,
+    I::IntoIter: ExactSizeIterator,
```
Kixunil, line 525: This is not needed because `bytes.len() * 2 == N`

src/display.rs
```diff
@@ -520,16 +522,21 @@ pub fn fmt_hex_exact_fn<I, const N: usize>(
...
     let mut encoder = BufEncoder::<N>::new();
-    encoder.put_bytes(bytes, case);
-    let encoded = encoder.as_str();
-
-    if let Some(precision) = f.precision() {
-        if encoded.len() > precision {
-            return f.pad_integral(true, "0x", &encoded[..precision]);
+    let bytes = bytes.into_iter();
+    let encoded = match f.precision() {
+        Some(p) if p < bytes.len() * 2 => {
```
Kixunil, line 530: Just use `p < N` here.

Kixunil (changes requested):
Please, just use `N`. Also the claim that it wouldn't be API-breaking is incorrect because the function is called from a public `fmt_hex_exact` macro which (arguably accidentally) accepts any type that has a `len` method and implements appropriate `IntoIterator`.

src/buf_encoder.rs
```diff
@@ -58,7 +58,7 @@ impl<const CAP: usize> BufEncoder<CAP> {
 
     #[inline]
     #[track_caller]
-    fn put_bytes_inner<I>(&mut self, bytes: I, case: Case)
+    pub(crate) fn put_bytes_inner<I>(&mut self, bytes: I, case: Case)
```
Kixunil, line 61: I don't think this is needed. Just calling `put_bytes` with an iterator works already.

src/display.rs
```diff
@@ -522,14 +524,18 @@ where
     I::Item: Borrow<u8>,
 {
     let mut encoder = BufEncoder::<N>::new();
-    encoder.put_bytes(bytes, case);
-    let encoded = encoder.as_str();
-
-    if let Some(precision) = f.precision() {
-        if encoded.len() > precision {
-            return f.pad_integral(true, "0x", &encoded[..precision]);
+    let bytes = bytes.into_iter();
```
Kixunil, line 527: This is not wrong but no longer needed since you can just call `.into_iter().take(n)` for the limited case. You may change it if you want to but it's not required.

src/display.rs
```diff
@@ -522,14 +524,18 @@ where
...
+    let encoded = match f.precision() {
+        Some(p) if p < N => {
+            let n = (p + 1) / 2;
```
Kixunil, line 530: Funny that the check above also happens to avoid overflow.

Kixunil (commented):
I'd rather keep our internal API clean, other than that it looks good.

apoelstra:
The CI failure is unrelated.

tcharding (approved):
ACK b036c1713ff5284afaacd6cf19588f7a936b143a

src/display.rs
```diff
@@ -587,6 +592,8 @@ mod tests {
             let dummy = Dummy([42; 32]);
             assert_eq!(dummy.to_string(), "2a".repeat(32));
             assert_eq!(format!("{:.10}", dummy), "2a".repeat(5));
+            assert_eq!(format!("{:.11}", dummy), "2a".repeat(5) + "2");
```
tcharding, line 595: TIL, I did not know that `String` implemented `ops::Add`.

apoelstra: Yes, and I hate it. `+` is commutative in basically every single mathematical context. I complained about this in the pre-1.0 days and said that (a) we should have a `++` operator for concatenation and was told that "this isn't Haskell", and (b) some muddied argument along the lines that `*` is non-commutative for matrices and nobody is bothered by that. IOW "fuck off nerd".

Anyway this is fine in test code but if we were using `+` on strings or vectors in real code I would have NACKed it.

apoelstra:
@Kixunil if you can re-review this (it's shorter now, and it wasn't that long originally) it'll be good to go.

src/display.rs
Kixunil: Daaamn! This annoyed me too and I was thinking this is one of the few things PHP got right.

Kixunil:
Crap, I have to run :(

Kixunil (approved):
ACK b036c1713ff5284afaacd6cf19588f7a936b143a

Issue #84, opened 2023-01-23
Re-implement `HexWriter`


tcharding:
Re-implement the `HexWriter` we removed in rust-bitcoin/rust-bitcoin#1572. See discussion thread for comments on improving the original by making it more generic.

https://github.com/rust-bitcoin/rust-bitcoin/pull/1572#issuecomment-1399507629

Make it generic over [`fmt::Write`](https://doc.rust-lang.org/std/fmt/trait.Write.html).

PR #109, opened 2024-09-13
Re-implement `HexWriter`

leo-nash:
Given a `T: fmt::Write`, `HexWriter` implements `std::io::Write` and writes the source bytes to its inner `T` as hex characters.

Fix: #84


src/display.rs
```diff
@@ -538,6 +538,43 @@ where
...
+where
+    T: core::fmt::Write,
+{
+    fn write(&mut self, buf: &[u8]) -> Result<usize, std::io::Error> {
+        let mut n = 0;
+        let table = self.1.table();
+        for byte in buf {
+            let pair = table.byte_to_hex(*byte);
+            if self.0.write_str(&pair).is_err() {
+                break;
```
Kixunil, line 569: This needs to return `Err(std::io::ErrorKind::Other.into())`.

Kixunil (changes requested):
Aside from error handling, this looks good.

src/display.rs
leo-nash: Thank you for the review. Help me understand:
- In [`std::io::Write::write`](https://doc.rust-lang.org/stable/std/io/trait.Write.html#tymethod.write) docs, they say _"If an error is returned then no bytes in the buffer were written to this writer."_ So here we would need to somehow "clear" the `T` upon returning an `std::io::Error`? Not sure of the correct approach to do this given `T: fmt::Write` alone.
- As an example, when `ArrayString::write_str` returns an `fmt::Error`, it is only an `arrayvec::CapacityError`. It seems to me this matches better to an `Ok(0)` return value from `std::io::Write` rather than a `std::io::Error` because _"It is not considered an error if the entire buffer could not be written to this writer."_ (from `std::io::Write` docs)
- Then there is this piece from [`core::fmt::Error`](https://doc.rust-lang.org/stable/core/fmt/struct.Error.html): _"This type does not support transmission of an error other than that an error occurred. This is because, despite the existence of this error, string formatting is considered an infallible operation."_

I can see how propagating `ErrorKind::Other` could be a nice way to tell people _"hey we don't know what happened, but `fmt::Write::write_str` returned an error."_ If we go with this approach, I would need to figure out a way to clear the "T" as described above.

Kixunil: Sadly, I think we can't follow the contract exactly but it should still work fine for consumers not relying on it (not doing any async operations). We should certainly document that it's imprecise.

I was originally thinking of also having `io::Write` -> `io::Write` but discarded the idea because I thought people could just use `fmt2io`. I was just thinking of adding it anyway but there's still an edge case when the underlying writer writes only one hex digit - we can't return 0.5.

apoelstra: Maybe if `n > 0` we should just truncate (as is done) to indicate "partial write succeeded". But if `n == 0` (i.e. the user tried to continue a partial write or called `write_all` which'll do it for him) then we return the actual error, which we will hit immediately.

I think this upholds the io::Write contract without requiring any exotic tricks.

We should have comments clearly explaining what we're doing here.

leo-nash: > But if n == 0 (i.e. the user tried to continue a partial write or called write_all which'll do it for him) then we return the actual error, which we will hit immediately.

I went with this solution in eaea55bfbe440fb659e64ef617ddec20a2010539 below - we return the error if n == 0, **and** the provided buffer is not empty.

Kixunil: @apoelstra I guess it can work but it will have worse performance. We have to write digit-by-digit and buffer the unwritten second digit if any. Then report success on number of bytes and implement `flush()` to attempt writing the buffered byte.

src/display.rs
```diff
@@ -538,6 +538,51 @@ where
...
+    /// # Errors
+    ///
+    /// If no bytes could be written to this `HexWriter`, and the provided buffer is not empty,
+    /// returns [`std::io::ErrorKind::Other`], otherwise returns `Ok`.
+    fn write(&mut self, buf: &[u8]) -> Result<usize, std::io::Error> {
+        let mut n = 0;
+        let table = self.1.table();
+        for byte in buf {
+            let pair = table.byte_to_hex(*byte);
+            if self.0.write_str(&pair).is_err() {
```
Kixunil, line 572: This is writing two-hex digits (chars), we have no way of knowing if both were written on error or not. So the only option is to write the first one and buffer the second one if there was an error.

src/display.rs
```diff
@@ -538,6 +538,51 @@ where
...
+            }
+            n += 1;
+        }
+        if n == 0 && !buf.is_empty() {
+            Err(std::io::ErrorKind::Other.into())
+        } else {
+            Ok(n)
+        }
+    }
+    fn flush(&mut self) -> Result<(), std::io::Error> { Ok(()) }
```
Kixunil, line 583: This will need to flush the buffered byte.

src/display.rs
```diff
@@ -538,6 +538,51 @@ where
...
+}
```
Kixunil, line 584: Note that if we implement byte-by-byte writing we really should override `write_all` to provide a more efficient implementation.

Kixunil (changes requested):

src/display.rs
apoelstra: We could just "buffer" a flag saying to return an error on the next call.

leo-nash: > This is writing two-hex digits (chars), we have no way of knowing if both were written on error or not.

I went with `write_str` to tell `T: fmt::Write`: ***either** write 2 chars, **or** return an error, **don't** write only one*. 

Upon error, I assume that zero characters have been written.

I understand this might be a shaky assumption, for `arrayvec::ArrayString` it holds up. Let me know what you think.

>  So the only option is to write the first one and buffer the second one if there was an error.

I am hesitant to use something like `write_char` to write character by character. If the first character write returns ok, but the second character write returns error, my writer has written half a byte, and has no way of telling `std::io::Write` exactly how many bytes it's written because of the unsigned integer return type (as poelstra has pointed out above).

I understand this could also be a problem with `write_str` method currently in use should my assumption about atomic 2 character writes not hold.

> We could just "buffer" a flag saying to return an error immediately on the next call.

Happy to do that yes. Is that the performance issue you've been referring to @Kixunil ?

PR #113, opened 2024-09-28
Re-implement `HexWriter`

tankyleo:
Given a `T: fmt::Write`, `HexWriter` implements `std::io::Write` and writes the source bytes to its inner `T` as hex characters.

Fix: #84

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -30,7 +30,7 @@ use alloc::string::String;
 use core::borrow::Borrow;
 use core::fmt;
 
-use super::Case;
+use super::{Case, Table};
 use crate::buf_encoder::BufEncoder;
 
 /// Extension trait for types that can be displayed as hex.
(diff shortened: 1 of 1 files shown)


tankyleo:
Sorry for the disruption.

Last time, we discussed possibly storing additional state in `HexWriter` so that in the case where:
1) we successfully write *some* bytes,
2) but `write_str` eventually returns an error on some nth byte

we put `HexWriter` in an error state so that the next call to `std::io::Write::write` immediately returns an error instead of trying a `write_str` again.

@Kixunil raised performance concerns, I hope this is what he refers to.

apoelstra:
The previous PR for this was #109 which mysteriously has been erased from history. (But it is in my browser history so I know it was there.)

I guess I need to start locally archiving github threads..

anyway c1f14567e47b450516f4b830ed1bb4ed8d29f93a looks good to me. As discussed in the previous thread, if the formatter somehow fails halfway through writing a byte, we don't detect this ... but there is also no way in the io::Write API for us to signal this, so I think we'll have to live with it.

...Actually, on further thought, we *could* signal it by setting a flag in the `HexWriter` that the user could check. But I think this is not worth the API complexity. In practice if users worry about half-written bytes and want to respond to it other than by giving up, they should probably write their own custom converter. It's not too hard.

apoelstra (approved):
ACK c1f14567e47b450516f4b830ed1bb4ed8d29f93a successfully ran local tests

tankyleo:
> As discussed in the previous thread, if the formatter somehow fails halfway through writing a byte, we don't detect this ... but there is also no way in the io::Write API for us to signal this, so I think we'll have to live with it.

My hope is that by using `std::fmt::Write::write_str`, it tells the formatter, *either* write the byte, *or* fail, *don't* write half a byte.

It works in the case where `arrayvec::ArrayString` is the formatter, but I am not sure about other formatters.

tcharding (approved):
ACK c1f14567e47b450516f4b830ed1bb4ed8d29f93a

apoelstra:
Let's give @Kixunil two weeks on this (from when the PR was opened) because he had some strong opinions on the last thread.

tcharding:
@tankyleo do you want this one or #117 to go in first?

tankyleo:
Rebase:
- Encoding performance PR broke this PR, fix it.
- Convert `HexWriter` to a struct, to make the style more consistent with `BufEncoder` and `BytesToHexIter`.

apoelstra (approved):
ACK fe7ced86022a84d4babd335992b60db93057e01c successfully ran local tests

tcharding (approved):
ACK fe7ced86022a84d4babd335992b60db93057e01c

PR #114, opened 2024-10-06
Fix `Display` width and precision of `DisplayByteSlice`

### LINKED ISSUE #70: `max` is divided by 2

### LINKED ISSUE #83: `DisplayHex` impls have different behaviours for different types
`DisplayArray`, `DisplayByteSlice`, and `fmt_hex_exact` each behave differently, this is surprising.

For example, how they handle `Formatter::precision`.

tankyleo:
Fixes #70

Also joins the implementation of `Display` for `DisplayByteSlice` and `DisplayArray`.

Partially addresses #83

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -166,8 +166,8 @@ impl<'a> DisplayByteSlice<'a> {
 
         let pad_right = if let Some(width) = f.width() {
             let string_len = match f.precision() {
-                Some(max) if self.bytes.len() * 2 > (max + 1) / 2 => max,
-                Some(_) | None => self.bytes.len() * 2,
+                Some(max) => core::cmp::min(max, self.bytes.len() * 2),
+                None => self.bytes.len() * 2,
             };
 
             if string_len < width {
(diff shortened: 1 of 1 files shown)


tankyleo:
No rush here.

- Tobin, I added you as co-author to eb1599b, how does that sound ? You had worked on this previously, the difference here is that this honors odd-length precision.
- The display of `fmt_hex_exact` is not yet consistent with `DisplayByteSlice` and `DisplayArray`. As a follow-up to this PR, we could have `internal_display` be written in terms of a `T: IntoIterator`, and have `fmt_hex_exact_fn` call `internal_display`

tcharding:
> Tobin, I added you as co-author to https://github.com/rust-bitcoin/hex-conservative/commit/eb1599b63a656f99f0542eddb7e76f6125399866, how does that sound ? You had worked on this previously, the difference here is that this honors odd-length precision.

No attribution required, I'll do review for you though :)

apoelstra (approved):
ACK addb8046d0758047925c55b6fe72e69a731c2761 successfully ran local tests

tcharding (approved):
ACK addb8046d0758047925c55b6fe72e69a731c2761

Issue #115, opened 2024-10-11
Are `hex_chars_to_byte` and `FromHex` optimized enough?


RCasatta:
It seems to me to much time is spent on this functions in my app, I think they are good candidate for performance optimization

![image](https://github.com/user-attachments/assets/40b79ebd-84f2-4247-8087-7359f18ebf36)

tcharding:
Interesting, here is `hex_chars_to_bytes`:

```rust
fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, InvalidCharError> {
    let hih = (hi as char).to_digit(16).ok_or(InvalidCharError { invalid: hi })?;
    let loh = (lo as char).to_digit(16).ok_or(InvalidCharError { invalid: lo })?;

    let ret = (hih << 4) + loh;
    Ok(ret as u8)
}
```

And here is `to_digit` (from `$RUST_SRC_PATH/core/src/char/methods.rs`:

```rust
    #[inline]
    pub const fn to_digit(self, radix: u32) -> Option<u32> {
        // If not a digit, a number greater than radix will be created.
        let mut digit = (self as u32).wrapping_sub('0' as u32);
        if radix > 10 {
            assert!(radix <= 36, "to_digit: radix is too high (maximum 36)");
            if digit < 10 {
                return Some(digit);
            }
            // Force the 6th bit to be set to ensure ascii is lower case.
            digit = (self as u32 | 0b10_0000).wrapping_sub('a' as u32).saturating_add(10);
        }
        // FIXME: once then_some is const fn, use it here
        if digit < radix { Some(digit) } else { None }
    }
```

I don't see any obvious bottle neck?

apoelstra:
We should check in godbolt and/or add some benchmarks. The existing code looks pretty close to optimal (in particular, after inlining and eliminating branches on the constant `radix`, that giant `to_digit` function will be super small), so to improve it we're going to need to look more closely. An obvious thing to try is to just reimplement `to_digit` in terms of `u8`s. Another good-to-try would be to use a lookup table, but that's not obviously faster than just using `wrapping_sub`s and assumptions about ASCII ordering.

This recent blog post is worth reading https://rundevelopment.github.io/blog/fast-unorm-conversions (though in his case, his original code has `f64` divisions so ofc he was able to get a huge speedup).

apoelstra:
[...]

PR #117, opened 2024-10-13
Encoding performance improvements

tankyleo:
Make significant performance improvements to `BytesToHexIter` and `DisplayHex`, see individual commits for details on the exact performance improvements.

Also deletes `fn byte_to_hex_chars` as it repeats the functionality offered by `Table::byte_to_hex`.

### DIFF
--- a/src/buf_encoder.rs
+++ b/src/buf_encoder.rs
@@ -38,7 +38,10 @@ impl<const CAP: usize> BufEncoder<CAP> {
     #[inline]
     #[track_caller]
     pub fn put_byte(&mut self, byte: u8, case: Case) {
-        self.buf.push_str(&case.table().byte_to_hex(byte));
+        let hex_chars: [u8; 2] = case.table().byte_to_hex(byte);
+        // SAFETY: Table::byte_to_hex returns only valid ASCII
+        let hex_str = unsafe { core::str::from_utf8_unchecked(&hex_chars) };
+        self.buf.push_str(hex_str);
     }
 
     /// Encodes `bytes` as hex in given `case` and appends them to the buffer.
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -10,6 +10,7 @@ use core::str;
 use std::io;
 
 use crate::error::{InvalidCharError, OddLengthStringError};
+use crate::Table;
 
 /// Convenience alias for `HexToBytesIter<HexDigitsIter<'a>>`.
 pub type HexSliceToBytesIter<'a> = HexToBytesIter<HexDigitsIter<'a>>;
@@ -202,6 +203,8 @@ where
     iter: I,
     /// The low character of the pair (high, low) of hex characters encoded per byte.
     low: Option<char>,
+    /// The byte-to-hex conversion table.
+    table: &'static Table,
 }
 
 impl<I> BytesToHexIter<I>
@@ -210,7 +213,7 @@ where
     I::Item: Borrow<u8>,
 {
     /// Constructs a new `BytesToHexIter` from a byte iterator.
-    pub fn new(iter: I) -> BytesToHexIter<I> { Self { iter, low: None } }
+    pub fn new(iter: I) -> BytesToHexIter<I> { Self { iter, low: None, table: &Table::LOWER } }
 }
 
 impl<I> Iterator for BytesToHexIter<I>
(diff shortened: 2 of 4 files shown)


tankyleo:
This breaks #113, so I mark it as draft until #113 is merged, happy to be patient.

Please see https://github.com/tankyleo/hex-benchmarks for how I got the "Estimated Cycles" numbers in the commit messages.

Thank you for your time and feedback.

src/buf_encoder.rs
```diff
@@ -38,7 +38,10 @@ impl<const CAP: usize> BufEncoder<CAP> {
     #[inline]
     #[track_caller]
     pub fn put_byte(&mut self, byte: u8, case: Case) {
-        self.buf.push_str(&case.table().byte_to_hex(byte));
+        let byte = case.table().byte_to_hex(byte);
+        // SAFETY: Table::byte_to_hex returns only valid ASCII
+        let byte = unsafe { core::str::from_utf8_unchecked(&byte) };
+        self.buf.push_str(byte);
```
tcharding, line 44:
```suggestion
        let hex_bytes = case.table().byte_to_hex(byte);
        // SAFETY: Table::byte_to_hex returns only valid ASCII
        let hex_str = unsafe { core::str::from_utf8_unchecked(&hex_chars) };
        self.buf.push_str(hex_str);
```

src/iter.rs
```diff
@@ -221,9 +224,9 @@ where
                 Some(c)
             }
             None => self.iter.next().map(|b| {
-                let (high, low) = byte_to_hex_chars(b);
-                self.low = Some(low);
-                high
+                let [high, low] = self.table.byte_to_hex(b);
+                self.low = Some(low as char);
```
tcharding, line 228: `char::from(low)` would be more clear here because makes it explicit that no data is being lost because it is infallible. As it is, the reader has to look up what type `low` is and then look up (or know) what casting that type to a `char` does.

tcharding: Some below for `high`.

tcharding (commented):
Cheers, looks good. The second patch I'm ambivalent on, we typically try to avoid casts but in that case its super obvious what is going on. The `char` casting would be better done with `char::from` IMO.

src/buf_encoder.rs
tankyleo: `hex_chars` would be `hex_bytes` here, but overall I agree thank you !

tankyleo:
> The second patch I'm ambivalent on, we typically try to avoid casts but in that case its super obvious what is going on. The char casting would be better done with char::from IMO.

I understand. Upon further thought, I'd be happy to drop the casts, and use `from` everywhere.

tankyleo:
Rebase:
- Drop all casts, replace with `from`.
- Rename variables for clarity.
- Name the return type explicitly for clarity.

apoelstra (approved):
ACK b659ecaf3973c4e52870f64f1d7bcab65b4fab0c successfully ran local tests

tcharding (approved):
ACK b659ecaf3973c4e52870f64f1d7bcab65b4fab0c

tcharding:
Nice, thanks. Did your github username change? I got confused seeing patches with a familiar name/email address for the author with a different username.

tankyleo:
I mark it as ready in case you'd like to merge this before #113

tcharding:
You can nominate your merge order preference if you want to.

tankyleo:
@tcharding your call. I have a preference for #117 first. It seems to me kix might want to review #113 before it goes in, as he had some concerns in a prior PR.

tcharding:
He has had two weeks already I believe. @apoelstra is the merger round here though so he will take care of the order. Cheers.

tcharding (approved):
ACK 29fb0a3bc53e0b39fc0a9f85fbb3f2662797b9fc

apoelstra:
Nice. Will do this one.

cc @RCasatta you might be interested in trying this on your profiling.

apoelstra (approved):
ACK 29fb0a3bc53e0b39fc0a9f85fbb3f2662797b9fc successfully ran local tests

RCasatta:
> cc @RCasatta you might be interested in trying this on your profiling.

In https://github.com/rust-bitcoin/hex-conservative/issues/115 I am interested in the opposite conversion, but thanks a lot for citing me, very nice perf improvement and TIL iai

PR #119, opened 2024-10-13
Restrict `BufEncoder` to uniform case encoding

tankyleo:
The API now enforces that all hex strings encoded with `BufEncoder` contain characters of the same case.

PR #121, opened 2024-10-15
Encapsulate unsafe code inside the table module

tankyleo:
The table module now returns only `[char; 2]` or `&str`, at the caller's choice.

Once `Table::byte_to_str` is called, the destination buffer cannot be written to until the scope of the returned reference ends.

### DIFF
--- a/src/buf_encoder.rs
+++ b/src/buf_encoder.rs
@@ -38,9 +38,8 @@ impl<const CAP: usize> BufEncoder<CAP> {
     #[inline]
     #[track_caller]
     pub fn put_byte(&mut self, byte: u8, case: Case) {
-        let hex_chars: [u8; 2] = case.table().byte_to_hex(byte);
-        // SAFETY: Table::byte_to_hex returns only valid ASCII
-        let hex_str = unsafe { core::str::from_utf8_unchecked(&hex_chars) };
+        let mut hex_chars = [0u8; 2];
+        let hex_str = case.table().byte_to_str(&mut hex_chars, byte);
         self.buf.push_str(hex_str);
     }
 
--- a/src/display.rs
+++ b/src/display.rs
@@ -152,7 +152,7 @@ fn internal_display(bytes: &[u8], f: &mut fmt::Formatter, case: Case) -> fmt::Re
         Some(max) if bytes.len() > max / 2 => {
             write!(f, "{}", bytes[..(max / 2)].as_hex())?;
             if max % 2 == 1 {
-                f.write_char(char::from(case.table().byte_to_hex(bytes[max / 2])[0]))?;
+                f.write_char(case.table().byte_to_chars(bytes[max / 2])[0])?;
             }
         }
         Some(_) | None => {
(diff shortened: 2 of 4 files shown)


tankyleo:
Hello Andrew, I share your concern about having redundant `unsafe`'s around codebase thank you for raising it.

I try to do the encapsulation here.

This breaks #119 , so we could merge after it if we do end up merging this.

tcharding:
The patch looks good but the terminology is confusing, specifically your use of the word 'flow' - do you mean scope or some function call, sorry I can't work it out.

tankyleo:
> The patch looks good but the terminology is confusing, specifically your use of the word 'flow' - do you mean scope or some function call, sorry I can't work it out.

Thank you, yes I mean the scope of the returned reference. I took that term flow from a book, but now I see that it is too specific, happy to switch to using the word scope.

tankyleo:
People would have to keep in mind that the scope of a reference ends when last used - not at the end of the function or block.

tankyleo:
Rebase:
- Add some tests
- Reword the documentation

tankyleo:
Hope this reads better Tobin, let me know.

tankyleo:
Rebase: fix the docs in a couple more places.

apoelstra:
In 4df89775da00b76b8b2953886f64e7bc6058173f:

This is possibly even more confusing. You say "`hex_chars` cannot be written to while `hex_str` is in scope" but there are only 3 lines in which `hex_chars` is even alive, and one of those lines is writing to it! I guess you're saying that the lifetime of the `&mut hex_chars` is the same as the lifetime of `hex_str`, but this isn't relevant to a reader of the code and doesn't need to be mentioned.

tankyleo:
I see, my bad here.

Here is what I mean to say:

This is OK:
```rust
        let mut hex_chars = [0u8; 2];
        let hex_str = case.table().byte_to_str(&mut hex_chars, byte);
        self.buf.push_str(hex_str);
        hex_chars = [0u8; 2];
```

This is not OK:
```rust
        let mut hex_chars = [0u8; 2];
        let hex_str = case.table().byte_to_str(&mut hex_chars, byte);
        hex_chars = [0u8; 2];
        self.buf.push_str(hex_str);
```

I think I'm causing unnecessary confusion here - I'm happy to drop all the comments that talk about this idea.

apoelstra:
>I'm curious how you would describe this idea.

How does the borrow checker describe it?

>I think I'm causing unnecessary confusion here - I'm happy to drop all the comments that talk about this idea.

I think we should drop it. The comment appears to be saying that if somebody were to write alternate, wrong, code, then the result wouldn't compile.

tankyleo:
Rebase: delete confusing comments.

tankyleo:
> How does the borrow checker describe it?
```
error[E0506]: cannot assign to `hex_chars` because it is borrowed
  --> src/buf_encoder.rs:43:9
   |
42 |         let hex_str = case.table().byte_to_str(&mut hex_chars, byte);
   |                                                -------------- `hex_chars` is borrowed here
43 |         hex_chars = [0u8; 2];
   |         ^^^^^^^^^^^^^^^^^^^^ `hex_chars` is assigned to here but it was already borrowed
44 |         self.buf.push_str(hex_str);
   |                           ------- borrow later used here

For more information about this error, try `rustc --explain E0506`.
```
> I think we should drop it. The comment appears to be saying that if somebody were to write alternate, wrong, code, then the result wouldn't compile.

Yes.

apoelstra (commented):
On 6cd4e456c61cb5576015b19026a8a8abee949834 successfully ran local tests

apoelstra (approved):
ACK 6cd4e456c61cb5576015b19026a8a8abee949834; successfully ran local tests

tcharding (approved):
ACK 6cd4e456c61cb5576015b19026a8a8abee949834

Issue #100, opened 2024-08-19
Return `char` instead of `u8` from `InvalidCharError::invalid_char`


Kixunil:
If the string contains a multi-byte UTF-8 char then the error returns confusing value unsuitable for the user.

This change is kinda breaking but we can phase it slowly: add `invalid_char_human(&self) -> char` method (please invent a better name!), implement `char` -> `u8` conversion inside `invalid_char` by encoding the first byte of the char and deprecate it.

I suspect this may need a bunch of changes to be able to get the entire char though.

apoelstra:
We had a similar bug in rust-bech32 in our new `Hrp::from_display` API. Fortunately we caught it before releasing.

Agreed with your strategy. Ideas for the new method name

* `invalid`
* `invalid_character`
* `character`
* `char_` (lol)

Kixunil:
`character` sounds least bad to me.

tcharding:
I had a play with this which led me to the view that

- Adding checks for utf-8 slows down parsing (because without some complex trickery we need to parse the input string twice)
- Giving a nice error message for utf-8 adds either complexity to the lib or a perf hit or both
- Parsing utf-8 (non ascii) characters is an edge case not the normal case
- We should not have a perf hit or unnecessary complexity for an edge case

I hacked up a `validate_hex_string` function but it is so trivial and undiscoverable that users will likely just get stuck debugging then write their own when they think of it.

```rust

/// Validates that input `string` contains valid hex characters.
pub fn validate_hex_string(s: &str) -> Result<(), NonHexDigitError> {
    for c in s.chars() {
        if !c.is_ascii_hexdigit() {
            return Err(NonHexDigitError { invalid: c});
        }
    }
    Ok(())
}
```

Perhaps a middle ground would be to add this to crate docs under some sort of `# Debugging` section?

apoelstra:
UTF-8 is designed so that as soon as you hit a non-ASCII byte you know that it's the start of a non-ASCII character. So we shouldn't need to use the slow `chars` iterator or anything. Just, when we hit a non-ASCII character then we need to maybe do some slow stuff to extract the full character that we've run into. (Taking `s[pos..].chars().next().unwrap()` would work for example, if we know the byte position of the bad character. And knowing this won't slow us down at all.)

tcharding:
That sounds feasible, I'll have a play with it. Thanks

Kixunil:
[...]

PR #124, opened 2024-11-09
Add `InvalidCharError::char`

tankyleo:
Returns a `char` if the invalid byte is ascii, otherwise returns the raw `u8`.

Deprecates `InvalidCharError::invalid_char`.

Partially addresses #100

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -80,14 +80,29 @@ pub struct InvalidCharError {
 
 impl InvalidCharError {
     /// Returns the invalid character byte.
+    #[deprecated(note = "Use `InvalidCharError::char` instead.")]
     pub fn invalid_char(&self) -> u8 { self.invalid }
     /// Returns the position of the invalid character byte.
     pub fn pos(&self) -> usize { self.pos }
+    /// Returns the invalid character.
+    ///
+    /// # Errors
+    ///
+    /// Returns the raw byte if it is not valid ascii.
+    pub fn char(&self) -> Result<char, u8> {
+        match self.invalid.is_ascii() {
+            true => Ok(char::from(self.invalid)),
+            false => Err(self.invalid),
+        }
+    }
 }
 
 impl fmt::Display for InvalidCharError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "invalid hex char {} at pos {}", self.invalid, self.pos)
+        match self.char() {
+            Ok(c) => write!(f, "invalid hex char {} at pos {}", c, self.pos),
+            Err(u) => write!(f, "invalid hex char {} at pos {}", u, self.pos),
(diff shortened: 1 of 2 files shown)


tankyleo:
This does not handle multi-byte characters yet.

- All `HexToBytesIter` needs is a `T: Iterator<Item = [u8; 2]> + ExactSizeIterator`, as shown in `HexToBytesIter::from_pairs`. I am reluctant to add more constraints on the `T` to peek for multi-byte characters in an error scenario.
- Even if we did handle multi-byte characters, not all of them would result in a nice visual picture. Is it worth the trouble then to handle multi-byte chars? I am not convinced.

tankyleo:
Is `HexToBytesIter::from_pairs` meant to be part of the public API ? Right now it is a `pub fn` inside a private module.

apoelstra:
>This does not handle multi-byte characters yet.

In this case the returned `char` is incorrect and we can't move forward with this approach.

tankyleo:
> In this case the returned char is incorrect and we can't move forward with this approach.

At the moment we only return a char if it is a single-byte character right ?

I agree that multi-byte chars would need more work.

apoelstra:
> At the moment we only return a char if it is a single-byte character right ?

No, we currently do not return a `char` under any circumstances, because the error type does not contain enough information to return a `char`.

tankyleo:
Right I was talking about this current commit here all good.

apoelstra:
Oh, I see! I misunderstood your initial approach.

This is clever but I am still going to concept NACK it. If users want to do this (treat ASCII specially) it is easy for them to do it themselves -- but this change forces them to do so, which adds extra code complexity.

tankyleo:
Sounds great I agree.

Was there another approach you had in mind for #100 ? If I understand you correctly we would not address this issue.

apoelstra:
What I would imagine is that if we read an invalid u8, we attempt to continue reading u8s until we've parsed a complete UTF-8 codepoint. (We would not attempt to deal with combining characters or anything like that because Rust itself makes it hard.)

And if we fail *that*...hmm. Maybe we would want to fall back to your API anyway?

tcharding:
@apoelstra we chatted about this, bump so you can update your position if you want to.

apoelstra:
Yeah -- I think we don't want to target this issue *at all* now for 1.0. If there is some noninvasive non-performance-affecting way to do it, that's great, but I don't believe there is.

Meanwhile, if a user has a hex string with a bad character in it, and wants to learn more specifically what that character is (and where it is, etc), it's not hard to do this without help from the crate. (`s.chars_indices().find(|(_, x)| !x.is_ascii())` will do it for example).

tankyleo:
Sounds good, I close this PR.

Whenever appropriate, I let you close #100 as well.

PR #132, opened 2024-12-29
Update nightly-version file

tcharding:
The `nightly-version` file is 3 months old. We don't really have an update policy but it seems like 3 months is old enough.

Grab the current version from `rust-bitcoin`.


apoelstra:
Nice, thanks! Yeah, we should update as often as we're willing to tolerate.

apoelstra (approved):
ACK 0098582d3fc38cc07c9c09ac64f552e5d731cd24; successfully ran local tests; thanks!

Issue #130, opened 2024-12-29
Fix warning when building with `--no-default-features`


tcharding:
```bash
   Compiling hex-conservative v0.3.0 (/home/tobin/build/github.com/tcharding/hex-conservative/master)
warning: unused import: `Table`
  --> src/display.rs:33:19
   |
33 | use super::{Case, Table};
```

PR #133, opened 2024-12-29
Feature gate import

tcharding:
`Table` is only used if `test` or `std` - feature gate it.

Fix: #130


apoelstra (approved):
ACK 64657dac9a62bee916468726952a4243c2765592; successfully ran local tests; lgtm

PR #134, opened 2025-01-02
Make `write_err` macro private

tcharding:
This is a helper macro, it should never have been public.


apoelstra (approved):
ACK 56ffb401ef4733f0c749f11b225355793c007803; successfully ran local tests

PR #135, opened 2025-01-09
Make `InvalidCharError` support UTF-8

tcharding:
Currently the `InvalidCharError` error returns a `u8` if an invalid hex character is found. If that invalid character happened to be a `char` that encodes as UTF-8 using more than one byte (i.e. not ASCII) then this means the `u8` returned is not the same as the invalid character in the input string.

Add support for invalid hex characters that fall outside of a single ASCII byte.

Because there are so many edge cases, this is a best effort. I'm not claiming to have gotten this right for all input strings. We default to just returning the byte.

Currently a 4-byte encoded character that falls on an odd position in the hex string is not correctly handled but this patch is definitely an improvement none the less.

Fix: #100

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -97,28 +97,165 @@ impl From<OddLengthStringError> for ToBytesError {
     fn from(e: OddLengthStringError) -> Self { Self::OddLengthString(e) }
 }
 
-/// Invalid hex character.
+/// UTF-8 character is not an ASCII hex character.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct InvalidCharError {
-    pub(crate) invalid: u8,
+    pub(crate) invalid: char,
     pub(crate) pos: usize,
 }
 
 impl InvalidCharError {
-    /// Returns the invalid character byte.
-    pub fn invalid_char(&self) -> u8 { self.invalid }
-    /// Returns the position of the invalid character byte.
+    /// Returns the invalid character.
+    pub fn invalid_char(&self) -> char { self.invalid }
+
+    /// Returns the position in the input string of the invalid character.
     pub fn pos(&self) -> usize { self.pos }
 }
 
 impl fmt::Display for InvalidCharError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "invalid hex char {} at pos {}", self.invalid_char(), self.pos())
+        write!(
+            f,
+            "char {} at pos {} is not a valid ASCII hex character",
+            self.invalid_char(),
+            self.pos()
+        )
     }
 }
 
 #[cfg(feature = "std")]
 impl std::error::Error for InvalidCharError {}
+
+/// Invalid ASCII encoding of a hex digit.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct InvalidDigitError {
+    /// The high digit in the pair e.g., for 'ab' this is 'a'.
+    pub(crate) hi: u8,
+
+    /// The high digit in the pair e.g., for 'ab' this is 'b'.
+    pub(crate) lo: u8,
+
+    /// `true` if the invalid byte is the `hi` one, `false` if its the `lo` one.
+    pub(crate) is_hi: bool,
+
+    /// Position of the stat of the byte pair in the input string.
+    pub(crate) pos: usize,
+}
+
+impl InvalidDigitError {
+    /// Returns the invalid byte.
+    pub fn invalid_byte(&self) -> u8 {
+        if self.is_hi {
+            self.hi
+        } else {
+            self.lo
+        }
+    }
+
+    /// Returns the character position in input string of the invalid byte.
+    pub fn pos(&self) -> usize {
+        if self.is_hi {
+            self.pos
+        } else {
+            self.pos + 1
+        }
+    }
+
+    pub(crate) fn into_invalid_char_error(
+        self,
+        // The next item from the iterator.
+        next: Option<Result<u8, InvalidDigitError>>,
+    ) -> InvalidCharError {
+        // Try to map the UTF-8 values in `v` to an invalid character.
+        fn error(v: &[u8], pos: usize) -> Option<InvalidCharError> {
+            core::str::from_utf8(v)
+                .map(|s| InvalidCharError { invalid: s.chars().next().unwrap(), pos })
+                .ok()
+        }
+
+        let rich_error = match next {
+            Some(res) => match res {
+                // The next two bytes happen to be valid ASCII.
+                Ok(byte) => {
+                    let (hi, lo) = crate::iter::byte_to_hex_digits(byte);
+                    if !self.is_hi {
+                        let vals = [self.lo];
+                        error(&vals, self.pos + 1).or_else(|| {
+                            let vals = [self.lo, hi];
+                            error(&vals, self.pos + 1).or_else(|| {
+                                let vals = [self.lo, hi, lo];
+                                error(&vals, self.pos + 1)
+                            })
+                        })
+                    } else {
+                        let vals = [self.hi];
+                        error(&vals, self.pos).or_else(|| {
+                            let vals = [self.hi, self.lo];
+                            error(&vals, self.pos).or_else(|| {
+                                let vals = [self.hi, self.lo, hi];
+                                error(&vals, self.pos).or_else(|| {
+                                    let vals = [self.hi, self.lo, hi, lo];
+                                    error(&vals, self.pos)
+                                })
+                            })
+                        })
+                    }
+                }
+                // The next two bytes happen to be invalid ASCII.
+                Err(e) =>
+                    if !self.is_hi {
+                        let vals = [self.lo, e.hi];
+                        error(&vals, self.pos + 1).or_else(|| {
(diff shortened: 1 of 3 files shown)


tcharding:
BTW I did not read up on the intricacies of UTF-8 encoding, I probably should have.

tcharding:
Oh and I spent waaay to long on this today. Hope it was worth it.

src/iter.rs
```diff
@@ -226,17 +239,29 @@ impl ExactSizeIterator for HexDigitsIter<'_> {}
...
 
-/// `hi` and `lo` are bytes representing hex characters.
+/// `hi` and `lo` are bytes that encode hex digits as ASCII.
 ///
-/// Returns the valid byte or the invalid input byte and a bool indicating error for `hi` or `lo`.
-fn hex_chars_to_byte(hi: u8, lo: u8) -> Result<u8, (u8, bool)> {
-    let hih = (hi as char).to_digit(16).ok_or((hi, true))?;
-    let loh = (lo as char).to_digit(16).ok_or((lo, false))?;
+/// Returns the valid byte or a bool indicating error for `hi` or `lo`.
+fn hex_digits_to_byte(hi: u8, lo: u8) -> Result<u8, bool> {
```
apoelstra, line 245: In ce28e9ba01cda818439b36108e072435715552fc:

I find the word "digit" when used to apply to a *number*, when we are working with a character encoding that contains actual digits, is extremely confusing.

src/iter.rs
```diff
@@ -226,17 +239,29 @@ impl ExactSizeIterator for HexDigitsIter<'_> {}
...
+
     Ok(ret as u8)
 }
 
+/// The inverse of `hex_digits_to_byte`.
+pub(crate) fn byte_to_hex_digits(byte: u8) -> (u8, u8) {
+    let hi = byte >> 4;
+    let lo = byte & 0x0f;
+
+    let hi = char::from_digit(hi.into(), 16).unwrap();
```
apoelstra, line 259: In ce28e9ba01cda818439b36108e072435715552fc:

This function is so badly named that I consider it a bug in stdlib. We should just do this manually by indexing into a static array containing `"0123456789abcdef"`. (I am not interested in filing bugs against stdlib to be told that "this is not Haskell" and I shouldn't complain that about words being used wrong. Let's just forget this function and move on with our lives.)

This would also have clearer semantics regarding capitalization, and to the extent there is any performance difference at all, likely to be faster.

src/error.rs
```diff
@@ -97,29 +97,166 @@ impl From<OddLengthStringError> for ToBytesError {
...
+        )
     }
 }
 
 #[cfg(feature = "std")]
 impl std::error::Error for InvalidCharError {}
 
+/// Invalid ASCII encoding of a hex digit.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct InvalidDigitError {
```
apoelstra, line 131: In ce28e9ba01cda818439b36108e072435715552fc:

BTW, here you use "digit" correctly to refer to a numeral character.

apoelstra: ...though the structure is still a little confusing since it actually holds *two* digits. But that's fine.

src/error.rs
```diff
@@ -97,29 +97,166 @@ impl From<OddLengthStringError> for ToBytesError {
...
+    /// `true` if the invalid byte is the `hi` one, `false` if its the `lo` one.
+    pub(crate) is_hi: bool,
+
+    /// Position of the stat of the byte pair in the input string.
+    pub(crate) pos: usize,
+}
+
+impl InvalidDigitError {
+    /// Returns the invalid byte.
+    pub fn invalid_byte(&self) -> u8 {
```
apoelstra, line 147: In ce28e9ba01cda818439b36108e072435715552fc:

Ok, here you have a error which holds two digits, which are supposed to comprise a byte but don't. You have a method called `invalid_byte` which returns *one of the two digits*. I think this name is confusing. It should be called `invalid_digit`.

src/error.rs
```diff
@@ -97,29 +97,166 @@ impl From<OddLengthStringError> for ToBytesError {
...
+                    error(&vals, self.pos).or_else(|| {
+                        let vals = [self.hi, self.lo];
+                        error(&vals, self.pos)
+                    })
+                },
+        };
+        rich_error.unwrap_or(InvalidCharError {
+            invalid: if self.is_hi { self.hi.into() } else { self.lo.into() },
+            pos: self.pos,
+        })
```
apoelstra, line 242: In ce28e9ba01cda818439b36108e072435715552fc:

This logic is really confusing. It seems like if the original "bad digit" was an ASCII byte then that byte can just be promoted to `char` and we're done. If not, we need to read more. So you could have an initial `if self.invalid_byte() < 128 { return InvalidCharError { invalid: self.invalid_byte().into(), pos: self.pos() }`. (And ofc `self.invalid_byte` should be renamed, see above.) So you can do an early return in this case and save yourself at least one level of nesting. (And also fix the bug here `self.pos` is wrong and you actually meant to return `self.pos()`.)

src/error.rs
```diff
@@ -97,29 +97,166 @@ impl From<OddLengthStringError> for ToBytesError {
...
+
+    pub(crate) fn into_invalid_char_error(
+        self,
+        // The next item from the iterator.
+        next: Option<Result<u8, InvalidDigitError>>,
+    ) -> InvalidCharError {
+        // Try to map the UTF-8 values in `v` to an invalid character.
+        fn error(v: &[u8], pos: usize) -> Option<InvalidCharError> {
+            core::str::from_utf8(v)
+                .map(|s| InvalidCharError { invalid: s.chars().next().unwrap(), pos })
```
apoelstra, line 172: In ce28e9ba01cda818439b36108e072435715552fc:

It seems like this function will return `None` in the case that you give it bad UTF-8 (for example, a non-ASCII character immediately followed by an ASCII character, which never happens in UTF-8 because it would make it ambiguous how to scan backward from the ASCII character to the nearest character boundary.)

But then below you seem to interpret `None` as "promote one of the bad digits in `self` to a `char`". Which, aside from being logically wrong, might not even result in a valid character.

apoelstra:
Done first pass on ce28e9ba01cda818439b36108e072435715552fc

src/iter.rs
tcharding: Amusingly I don't even know what `hi` and `lo` are from just looking at this diff and I worked on this code yesterday for hours. I thought 'symbol' instead of 'digit' but perhaps 'hex_ascii_value_to_byte` but the perhaps `to_byte` is actually better described by 'decode' because the two hex symbols encode a byte. 

Perhaps: `decode_hex_ascii_value_pair`

tcharding:
Thanks for the review man. Seems this will not only take a lot of dev time but also a lot of review time - is it definitely worth doing? Or should we consider a bandaid solution like an error message that says "invalid character is `<whatever>`, note that if the invalid character falls outside of the single byte ASCII range then this error will not include the actual character"

src/iter.rs
apoelstra: Yep, `ascii_value_pair` sounds good to me.

And yeah, `hi` and `lo` are not great... they are obviously the two nybbles of the byte, but are they ASCII values or numbers? Not easy to tell when we are using `u8` for both.

apoelstra:
I think it'd be totally fine to say "invalid character `c`" if `c` is a printable ASCII byte and "invalid byte `0xAB`" if `c` is not.

I wouldn't call it a "bandaid solution" to accurately describe the error condition without implementing a UTF-8 parser for the sake of providing more context (which might not even be valuable BTW because it assumes that the input came from a UTF-8 string that the user can see in a decoded form).

apoelstra:
Right -- I complained about exposing an API that returns `Option<char>`. I continue to think this is not worth the extra API surface (it's not clear what the user would expect -- a char if the bad byte was ASCII? if it was printable? if it was alphanumeric? -- and even if we get it right and this can be quickly understood by the author and all readers of the code, it barely saves any characters because the inline logic is so simple).

But for purposes of implementing `Display` on this error I think it's reasonable to special-case printable characters.

tcharding:
This PR is difficult to work on and difficult to review. I have little motivation for it. Can we 1.0 without it?

Kixunil:
Honestly, I'm very much unsure about the API of iterators because to support the chars we might be forced to change the API to better track if the backing bytes are guaranteed string. As such I think we should separate these in a `hex-conservative-unstable` crate for the time being. That would allow us to skip this for 1.0. Though I'm worried it may cause a dependency cycle.

Really, the only things we need stable are the error types (well, kinda LOL about char but if we don't expose it then it's fine) and since we can't provide sensible constructors for them we also need trivial decoding functions that return them.

tcharding:
Before doing this could consider adding some type that abstracts over a pair of hex characters e.g.,

```rust
struct HexAsciiPair {
   hi: u8,
   low: u8,
}
```

The `iter` module is not going to be public in the initial 1.0 release so this can wait.

Kixunil:
@tcharding we could have that but I suspect it'd be a hindrance to using `chunks_exact` (or `chunk_arrays` in the future). We could probably use `#[repr(C)]` and `unsafe` cast them without much trouble.

Hmm, now I'm thinking maybe we should re-implement `chunk_arrays` in `internals`, why didn't I think of that before?

PR #136, opened 2025-01-20
Make a start on verifying the API

tcharding:
Add infrastructure to get us started checking the API surface as we are doing for the leaf crates in `rust-bitcoin`.

Includes CI to check for API changes.


tcharding:
hmm, I thinking we should add this to CI - @apoelstra?

apoelstra:
Yeah.

Since you posted this Kix briefly appeared -- if he sticks around we might need to remove it, but it's worth having even for a couple days to help us auditing the existing API.

tcharding:
Yes, cc @Kixunil for veto rights on this PR.

tcharding:
> but it's worth having even for a couple days to help us auditing the existing API.

I also use it in `contrib/api.sh` which I find useful for the checklist work.

apoelstra (approved):
ACK 1761f1761335c61d873d8f03f845a1ded4d43c81; successfully ran local tests; lgtm! May want to split out the API test since that is an uncontroversial improvement

tcharding:
Can do, no sweat.

tcharding:
I cannot actually be arsed to argue over this so I'm putting it in draft.

Kixunil:
Can't you just upload the reports to some kind of pastebin/gist/whatever and have the bot post a comment linking it? It'd be far less annoying for contributors without affecting reviewers (other than the ability to conveniently comment on specific line).

tcharding:
Close, I'll keep this live on my own machine so there is no implication that I'm championing it still.

apoelstra:
> without affecting reviewers

Lol, going to some random web service instead of using `git diff` is extremely annoying and disruptive. Aside from the work that the "reviewers" would need to do to set this up to begin with.

But fine, Tobin and I will find some way to just directly share git commits without them touching this repo.

Kixunil:
How about we have a separate repository hosting just the API files and a bot to sync it? You can then use git workflow very similar to what you already have. You could even have a local alias to sync the two repositories as needed. Or maybe submodules/subtrees can do some magic there.

apoelstra:
If we were gonna do that then we'd use a separate git branch within this repo containing the API files, since we could then retain the same repo. But it would still be less convenient because the API files would not be tied to the commits that were used to generate them. We would need to write some querying and diff tools, and decide how to handle changes in the number of workspaces or the parameters to cargo-public-api.

I can give this a shot, but it will be some work to write and still less convenient than having the API diff show up alongside the code diff.

tcharding:
The whole point is no one can be fucked writing the tooling, we have a solution that works, is ready to go, and has provable helped Andrew and myself review and develop. By all means write a better solution but there are only so many hours in the day and mine are all full.

Kixunil:
> But it would still be less convenient because the API files would not be tied to the commits that were used to generate them

Easy solution, just put the commit ID into git message.

Anyway, I take your answer as ACK for new tooling but you don't want to bother writing it too much. The current solution works for you but doesn't for me but there's two of you, so I think it's fair for me to figure out the tooling. It's fine for me, I will write it at some point (unless it becomes incredibly cumbersome) and meanwhile you can keep using the current one.

apoelstra:
>Easy solution, just put the commit ID into git message.

This is not a sufficient solution. If you want to simulate deterministic data implied by the contents of a git commit, but not in the git commit, then you need to do git-annex-like tricks where you maintain a mapping from commits to data along with logs and integrity checks, and deal with what happens when different users have different views of this data, and rewrite a bunch of basic git tooling because there's indirection involved in simple stuff like `git diff`.

We probably don't need all that for this usecase (in particular it's *probably* okay if different people sometimes have different views of the API files), but these are the properties that we get for free by just putting the data into the commits.

Meanwhile, the cost for maintainers who don't want to run `./contrib/check-public-api.sh` is that they need to copy/paste some output from CI, only when they are changing the public API.

Kixunil:
> Meanwhile, the cost for maintainers who don't want to run `./contrib/check-public-api.sh` is that they need to copy/paste some output from CI, only when they are changing the public API.

Oh, true that I can just copy-paste... Well, we'll see how it goes.

PR #140, opened 2025-01-21
Deprecate the `test_hex_unwrap` macro

tcharding:
We are trying to reduce the API surface and maintenance burden of the `1.0` release.

This macro is a oneliner. Downstream users can just define the macro themselves if they want it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,6 +154,7 @@ mod table {
 ///
 /// For improved ergonomics import with: `use hex_conservative::test_hex_unwrap as hex;`
 #[macro_export]
+#[deprecated(since = "TBD", note = "use the one-liner `Vec::from_hex(hex).unwrap()` instead")]
 macro_rules! test_hex_unwrap (($hex:expr) => (<Vec<u8> as $crate::FromHex>::from_hex($hex).unwrap()));
 
 #[cfg(test)]


apoelstra:
Lol, "deemed unnecessary" sounds weirdly callous. Like we considered that people might be using this but decided that they're just little people who can be swept aside.

If you want to say this, we could use language like "reducing API surface for 1.0; this is a one-liner".

Though I would suggest something like "use the one-liner `Vec::from_hex(hex).unwrap()` instead".

Kixunil:
It feels odd that we had it in the first place. Especially since `hex_lit` is superior in most cases.

apoelstra:
Well, within our own crates we probably wanted to avoid the extra dep for the sake of a one-liner. Though it would only be a dev-dependency.

But yeah, no point in us maintaining parallel functionality if we're gonna do it poorly and increase our own macro API surface area. (The context of this is that we have gotten really frustrated with exported macros and are trying to remove them wherever their benefit is small.)

tcharding:
> Lol, "deemed unnecessary" sounds weirdly callous. 

You should have seen me gesture with my hand, all I was missing was one of those fur cape things kings of yore wore.

apoelstra (approved):
ACK 2777b2373dbb808d608f343bbe2fb9b057614c0f; successfully ran local tests

apoelstra:
Gonna merge -- I think Kix's comment is an implicit ACK.

PR #141, opened 2025-01-21
Derive `PartialOrd` and `Ord` for `Case`

tcharding:
Currently the `Case` enum does not implement `PartialOrd` and `Ord`. This is an unnecessary restriction.

Done as part of the Rust API guidelines checklist.

ref: https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,7 +78,7 @@ pub use self::{
 };
 
 /// Possible case of hex.
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub enum Case {
     /// Produce lower-case chars (`[0-9a-f]`).
     ///


apoelstra (approved):
ACK 8fc6aee6524d1c9deea0d5ffa1a057a5fec4ca59; successfully ran local tests

Kixunil:
I'm generally careful about adding ordering to weird things but I guess it's harmless here. Note that the thing is called "guidelines" not "hard rules", we can have small disagreements. (Also traits can be easily added later, so being conservative is not too bad.)

apoelstra:
@Kixunil we decided somewhere elsewhere that for non-error enums that don't have any "obvious" ordering that might confuse or surprise people, we'd just implement the alphabetical ordering and document this. The convenience benefit of being able to `derive(PartialOrd, Ord)` is pretty high.

One weird case was the `Denomination` type in `Amount`, where we ordered the denominations by size and fell back to alphabetical for denominations that were equal in magnitude. You might want to take a look at that in case you disagree with us.

apoelstra:
Oh -- this is actually the open PR https://github.com/rust-bitcoin/rust-bitcoin/pull/3934 -- so plenty of time for more comments.

Kixunil:
WHAT???!!! We have differenet enum variants for the same unit?! That sounds very broken. (If we don't have them then I agree with the ordering from the smallest.)

apoelstra:
>That sounds very broken.

Yeah, it was a surprise to me too! We should move discussion to the other PR.

tcharding:
Same comment as https://github.com/rust-bitcoin/rust-bitcoin/pull/3934#issuecomment-2606222984

PR #142, opened 2025-01-21
Add missing error re-exports

tcharding:
We should be re-exporting all the errors in this crate so downsteam can use `hex::FooError` and not have to do `hex::error::FooError`.

Add the missing error re-exports, with this applied all the error types from `error` are re-exported at the crate root.


apoelstra (approved):
ACK b7a93268d631a6a7002c38ff7a0f29805ef2f2a5; successfully ran local tests

PR #143, opened 2025-01-22
Test `impl_fmt_traits` in module scope

tcharding:
As part of C-ANYWHERE add a unit test for `impl_fmt_traits` in module scope. Just copy the existing function scope code.


apoelstra:
You may need to allow dead code here, or write a simple test case that actually tries to construct this, or something.

tcharding:
Allowed! Thanks

apoelstra (approved):
ACK d48cd48e36df7651b81ed637c1cbd31b00a623a0; successfully ran local tests

PR #144, opened 2025-01-22
Document and test `HexWriter` constructor

tcharding:
Thanks to a blanket impl of `Write` for `&mut W` we can pass in a mutable reference to `HexWriter::new` even though the function signature suggests it takes ownership of the writer.

Document this as suggested in the Rust API guidelines doc.

While we are at it add a unit test to prove it.

ref: https://rust-lang.github.io/api-guidelines/interoperability.html#c-rw-value

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -582,6 +582,8 @@ pub struct HexWriter<T> {
 impl<T> HexWriter<T> {
     /// Creates a `HexWriter` that writes the source bytes to `dest` as hex characters
     /// in the given `case`.
+    ///
+    /// Note even though we take ownership of the writer one can also call this with `&mut dest`.
     pub fn new(dest: T, case: Case) -> Self { Self { writer: dest, table: case.table() } }
     /// Consumes this `HexWriter` returning the inner `T`.
     pub fn into_inner(self) -> T { self.writer }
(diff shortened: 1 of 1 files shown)


apoelstra:
rustfmt is unhappy with this

tcharding:
Why do you idiots even use the formatter in CI ... oh wait

apoelstra (approved):
ACK 0f8d2038d3b337449c8d30dc15939330d4156de9; successfully ran local tests

src/display.rs
```diff
@@ -966,5 +970,12 @@ mod tests {
             writer.write_all(&vec[..]).unwrap();
             assert_eq!(writer.into_inner(), vec.to_lower_hex_string());
         }
+
+        #[test]
+        fn hex_writer_accepts_and_mut() {
+            let mut dest_buf = ArrayString::<64>::new();
+            let mut dest = HexWriter::new(&mut dest_buf, Case::Lower);
+            let _got = dest.write(b"some data").unwrap();
```
Kixunil, line 978: This test is actually testing that `std` still has the `&mut T` impl. I'm certain we don't need to test that. :)

tcharding: Correct. It also proves, probably unnecessarily like you say, that the docs I wrote in this PR are correct.

Issue #125, opened 2024-11-18
Make the crate conform to the API guidelines


tcharding:
# Rust API Guidelines Checklist

- **Naming** *(crate aligns with Rust naming conventions)*
  - [x] Casing conforms to RFC 430 ([C-CASE])
  - [x] Ad-hoc conversions follow `as_`, `to_`, `into_` conventions ([C-CONV])
  - [x] Getter names follow Rust convention ([C-GETTER])
  - [x] Methods on collections that produce iterators follow `iter`, `iter_mut`, `into_iter` ([C-ITER])
  - [x] Iterator type names match the methods that produce them ([C-ITER-TY])
  - [x] Feature names are free of placeholder words ([C-FEATURE])
  - [x] Names use a consistent word order ([C-WORD-ORDER])
- **Interoperability** *(crate interacts nicely with other library functionality)*
  - [x] Types eagerly implement common traits ([C-COMMON-TRAITS])
    - `Copy`, `Clone`, `Eq`, `PartialEq`, `Ord`, `PartialOrd`, `Hash`, `Debug`,
      `Display`, `Default`
  - [x] Conversions use the standard traits `From`, `AsRef`, `AsMut` ([C-CONV-TRAITS])
  - [x] Collections implement `FromIterator` and `Extend` ([C-COLLECT])
  - [x] Data structures implement Serde's `Serialize`, `Deserialize` ([C-SERDE])
  - [x] Types are `Send` and `Sync` where possible ([C-SEND-SYNC])
  - [x] Error types are meaningful and well-behaved ([C-GOOD-ERR])
  - [x] Binary number types provide `Hex`, `Octal`, `Binary` formatting ([C-NUM-FMT])
  - [X] Generic reader/writer functions take `R: Read` and `W: Write` by value ([C-RW-VALUE])
- **Macros** *(crate presents well-behaved macros)*
  - [x] Input syntax is evocative of the output ([C-EVOCATIVE])
  - [x] Macros compose well with attributes ([C-MACRO-ATTR])
  - [x] Item macros work anywhere that items are allowed ([C-ANYWHERE])
  - [x] Item macros support visibility specifiers ([C-MACRO-VIS])
  - [x] Type fragments are flexible ([C-MACRO-TY])
- **Documentation** *(crate is abundantly documented)*
  - [ ] Crate level docs are thorough and include examples ([C-CRATE-DOC])
  - [ ] All items have a rustdoc example ([C-EXAMPLE])
[...]

PR #145, opened 2025-01-24
Verify the API

tcharding:
Add an integration test module to prove the API surface of this crate. Done as part of #125.

- Patch 1: Derive `Debug`
- Patch 2: Add explicit trait bound to the `HexWriter`
- Patch 3: Add the test module

This was pulled out of #136

### DIFF
--- a/src/buf_encoder.rs
+++ b/src/buf_encoder.rs
@@ -19,6 +19,7 @@ use super::{Case, Table};
 /// This is an important building block for fast hex-encoding. Because string writing tools
 /// provided by `core::fmt` involve dynamic dispatch and don't allow reserving capacity in strings
 /// buffering the hex and then formatting it is significantly faster.
+#[derive(Debug)]
 pub struct BufEncoder<const CAP: usize> {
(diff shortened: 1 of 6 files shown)


tcharding:
It could easily be argued that patch 2 is unnecessary and out of scope. I can remove it, no stress, if there is any push back whatsoever.

apoelstra:
We should remove the second commit. In general you shouldn't put trait bounds on type signatures unless it's necessary to make the code to compile.

By doing this, every time somebody wants to hold a `HexWriter<T>` in some composite structure, they'll be forced to write `T: fmt::Write` on *that* structure, and then they'll be forced to write `T: fmt::Write` on every impl block for that structure even for methods that don't even touch the hexwriter, and you wind up with a ton of unnecessary noise.

tcharding:
Done, thanks for the explanation.

apoelstra (approved):
ACK 4bf330dbb7f24298cfd8d4e02975cf7e84bac20a; successfully ran local tests

PR #147, opened 2025-02-13
Add an functionally empty dyn compatibility test

tcharding:
In order to complete the API guidelines checklist add a test that is functionally empty but exists as documentation about the single public trait in this crate.

At a bare minimum it saves one wondering why the test wasn't added and having to check the trait, then look up what make dyn compatibility not possible ...

### DIFF
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -140,3 +140,11 @@ fn all_types_implement_send_sync() {
     assert_send::<Errors>();
     assert_sync::<Errors>();
 }
+
+#[test]
+fn dyn_compatible() {
+    // If this builds then traits are dyn compatible.
+    struct Traits {
+        // a: Box<dyn hex_conservative::FromHex>, // `from_hex()` returns `Self`.gs
+    }
+}


tests/api.rs
```diff
@@ -140,3 +140,11 @@ fn all_types_implement_send_sync() {
     assert_send::<Errors>();
     assert_sync::<Errors>();
 }
+
+#[test]
```
Kixunil, line 144: I think these kinds of tests should be effectively static asserts and can be written as `#[allow(dead_code)] fn is_foo_dyn_compatible(_: &dyn Foo) {}`. That of course implies we cannot write an empty test like this and have to use a comment.

Also it'd be useful to have an inverse test: that the other traits are *not* dyn compatible to ensure we don't accidentally make something dyn compatible and later change it back on accident again.

tcharding: I don't know how to write a test that assets some code does not build. Do you know? There are other things we can verify if we had such a way of writing tests.

I don't see the second being any better than the first?

```rust

#[test]
#[allow(dead_code)]
fn dyn_compatible() {
    // If this builds then traits are dyn compatible.
    struct Traits {
        a: Box<dyn amount::CheckedSum<Amount>>,
        // These traits are explicitly not dyn compatible.
        // b: Box<dyn amount::serde::SerdeAmount>,
        // c: Box<dyn amount::serde::SerdeAmountForOpt>,
        // d: Box<dyn parse::Integer>,
    }

#[test]
#[allow(dead_code)]
fn dyn_compatible() {
    fn is_checked_sum__dyn_compatible(_: &dyn amount::CheckedSum<Amount>) {}
    // These traits are explicitly not dyn compatible.
    // fn is_serde_amount_dyn_compatible(_: &dyn amount::serde::SerdeAmount) {}
    // fn is_serde_amount__for_opt_dyn_compatible(_: &dyn amount::serde::SerdeAmountForOpt) {}
    // fn is_integer_dyn_compatible(_: &dyn parse::Integer) {}
}
```

Kixunil: > Do you know?

I believe the only way is to have them in a separate directory and add `if cargo check -p foo; then exit 1; fi` in a bsh script/CI. Note though that this requires a crate for each individual test but maybe there's a hack for that. (E.g. pretend they are normal binaries and try to build only one?)

> I don't see the second being any better than the first?

You've overcomplicated it. You don't need `#[test]` or a second function. Just that single function thrown in the `test` module is enough. (Though, in practice maybe we should have a separate module for these just to slap `#[allow]` on all of them at once if we have more of them.)

tcharding: Yeah, that sounds better. Cheers

tcharding:
I'm going to take the `fn is_foo_dyn_compatible(_: &dyn Foo) {}` idea to the crates in `rust-bitcoin` but I'm not going to further pursue the 'prove traits are _not_ dyn compatible idea'

PR #149, opened 2025-02-17
Impl `From<Infallible>` for all Error types

### LINKED ISSUE #128: Error types do not impl `From<Infallible>`
We do this in `rust-bitcoin`. Should we be doing it here?

jamillambert:
Implement this conversion for all error types to make error handling easier.

Close #128

### DIFF
--- a/examples/custom.rs
+++ b/examples/custom.rs
@@ -3,6 +3,7 @@
 //! For a type where hex is the natural representation see `./hexy.rs`.
 //! To wrap an array see the `./wrap_array_*` examples.
 
+use core::convert::Infallible;
 use core::fmt;
 use core::str::FromStr;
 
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK da4dcdf3432d03d69e7905ce25a8cab897fba6e2; successfully ran local tests

examples/custom.rs
```diff
@@ -124,6 +125,10 @@ pub enum Error {
     InvalidStringFormat,
 }
 
+impl From<Infallible> for Error {
+    fn from(never: Infallible) -> Self { match never {} }
+}
```
Kixunil, line 130: I think examples should be exempt from adding things they don't use but I won't block this.

Kixunil (approved):
ACK da4dcdf3432d03d69e7905ce25a8cab897fba6e2

Issue #139, opened 2025-01-21
Do we really want to commit to the traits in `1.0`?


tcharding:
Since writing this crate our usage in `bitcoin` has evolved. We only use inherent functions to parse (`from_hex`). To encode we use  `Display`/`LowerHex` coupled with an inherent `to_hex` function. I believe we could seal the `FromHex` and `DisplayHex` traits and our usage in `bitcoin` would not be effected.

Do we really want to release and maintain a general purpose hex library or do we want to provide a library that is used by `rust-bitcoin`?

If its the latter we can seal up the traits and reduce the maintenance burden.

Said another way: this library really just provides encode/decode to/from vectors and arrays - a trait is required for the orphan rule but do we really want others to implement the traits?

tcharding:
Further evidence that this is not a general hex crate: we have a bunch of integer parsing stuff in `units` (e.g. `hex_32_unchecked`).

apoelstra:
Absolutely not. We tried "let's just delete the hex stuff" in 0.30 and it caused several people, including me, to avoid upgrading for the better part of a year or more. Then we had a huge stupid fight about calling the crate `bitcoin-hex` because it "wasn't a general hex crate" and we landed on the name we have.

> Do we really want to release and maintain a general purpose hex library

Yes.

>do we really want others to implement the traits?

Yes.

Kixunil:
We could've just had it unstable (not forever, just for a while) if not for the error types. I think we can still do it by copying (and converting) the error types, but do we want to?

apoelstra:
I think, if that's the cost of getting stability reasonably quickly, it'd be worth doing.

Kixunil:
Oh, maybe you're right since we can add `From` to them once both are stable which would make it less annoying than I thought. (But meanwhile we need `map_err(convert_hex_error)` in our codebase.)

tcharding:
>  I think we can still do it by copying (and converting) the error types, but do we want to?

We decided against this just a few weeks ago here: https://github.com/rust-bitcoin/hex-conservative/pull/129

It does seem meaningful that we are debating this again such a short time afterwards. (No I did not do that intentionall, this issue just struck me yesterday while I was working through the Rust API guidelines checklist.)

Really good timing for you to be back @Kixunil, I was getting increasingly nervous about 1.0'ing `hex` without you signing off on it.

apoelstra:
>It does seem meaningful that we are debating this again such a short time afterwards

[...]

PR #151, opened 2025-02-18
Remove `FromHex` from `hexy` example

tcharding:
The `hexy` example is specifically supposed to showcase a type for which the natural representation is hexadecimal. It does not need to implement `FromHex`.

Remove the `FromHex` impl and inline the logic into `FromStr`.

Done as part of #139


tcharding:
This is too small to do as a single PR - will do the whole issue in one PR.

PR #152, opened 2025-02-18
Seal all public traits

tcharding:
As part of 1.0 preparation we would like to reduce the API surface as much as possible.

Seal the `FromHex` and `DisplayHex` traits. These are the only pubic traits in this crate.

Close: #139


Kixunil (approved):
ACK 43135a5592ad3b4648de16f28f85710455729aeb

apoelstra:
It's separate from this PR, but why do we impl one trait for all `N: usize` and the other one for specific fixed array sizes?

apoelstra (approved):
ACK 43135a5592ad3b4648de16f28f85710455729aeb; successfully ran local tests

tcharding:
Oh just thought that your CI box might fall over now when it finds this merged out from under it.

tcharding:
> It's separate from this PR, but why do we impl one trait for all `N: usize` and the other one for specific fixed array sizes?

I took a quick look and its because of const generics not working how one would suppose (assuming you suppose they work how I suppose they should work)

```bash
error: generic parameters may not be used in const operations
   --> src/display.rs:339:38
    |
339 |     type Display = DisplayArray<'a, {LEN * 2}>;
    |                                      ^^^ cannot perform const operation using `LEN
```

PR #155, opened 2025-02-19
Enforce `BufEncoder` capacity check for evenness

### LINKED ISSUE #150: Even `BufEncoder` capacity not enforced
In `buf_encoder.rs:29` the following suggests that only an even capacity buffer is allowed.  But it is not enforced and it is possible to create an odd sized buffer in a test.
```rust 
const _CHECK_EVEN_CAPACITY: () = [(); 1][CAP % 2];
```
I assume the compiler is optimizing it away since it is not referenced anywhere.

jamillambert:
The check that `BufEncoder` capacity is even was intended to not compile if it was odd.  But the constant created for this was never used and the compiler optimised it away.  This allowed odd capacities to be used.

Use `_CHECK_EVEN_CAPACITY` in `new()` to ensure only an even capacity `BufEncoder` can be created.

### DIFF
--- a/src/buf_encoder.rs
+++ b/src/buf_encoder.rs
@@ -30,7 +30,11 @@ impl<const CAP: usize> BufEncoder<CAP> {
 
     /// Creates an empty `BufEncoder` that will encode bytes to hex characters in the given case.
     #[inline]
-    pub fn new(case: Case) -> Self { BufEncoder { buf: ArrayString::new(), table: case.table() } }
+    #[allow(clippy::let_unit_value)] // Allow the unit value of the const check
+    pub fn new(case: Case) -> Self {
+        let _ = Self::_CHECK_EVEN_CAPACITY;
+        BufEncoder { buf: ArrayString::new(), table: case.table() }
+    }
 
     /// Encodes `byte` as hex and appends it to the buffer.
     ///


src/buf_encoder.rs
```diff
@@ -30,7 +30,11 @@ impl<const CAP: usize> BufEncoder<CAP> {
 
     /// Creates an empty `BufEncoder` that will encode bytes to hex characters in the given case.
     #[inline]
-    pub fn new(case: Case) -> Self { BufEncoder { buf: ArrayString::new(), table: case.table() } }
+    #[allow(clippy::let_unit_value)] // Allow the unit value of the const check
```
Kixunil, line 33: Note: this lint is stupid. It makes something that was already explicit explicit.

Kixunil (approved):
ACK ec591b7f3f6371e8a58cdde72bded5858816f8ac

src/buf_encoder.rs
apoelstra: https://github.com/rust-lang/rust-clippy/issues/9080

apoelstra (approved):
ACK ec591b7f3f6371e8a58cdde72bded5858816f8ac; successfully ran local tests

PR #156, opened 2025-02-19
Add examples to BufEncoder

jamillambert:
An even capacity `BufEncoder` is now enforced.

Add a line to the rustdocs stating that the capacity must be even.

Add a rustdoc example to BufEncoder that shows the correct usage and also a case with an odd capacity that results in a compile failure.

### DIFF
--- a/src/buf_encoder.rs
+++ b/src/buf_encoder.rs
@@ -19,6 +19,25 @@ use super::{Case, Table};
 /// This is an important building block for fast hex-encoding. Because string writing tools
 /// provided by `core::fmt` involve dynamic dispatch and don't allow reserving capacity in strings
 /// buffering the hex and then formatting it is significantly faster.
+///
+/// The buffer has a fixed capacity specified when created. The capacity must be an even number since
+/// each byte is encoded as two hex characters.
+///
+/// # Examples
+/// ```
+/// # use hex_conservative::buf_encoder::BufEncoder;
+/// # use hex_conservative::Case;
+/// let mut encoder = BufEncoder::<4>::new(Case::Lower);
+/// encoder.put_byte(0xab);
+/// assert_eq!(encoder.as_str(), "ab");
+/// ```
+/// The following code doesn't compile because of odd capacity:
+/// ```compile_fail
+/// # use hex_conservative::buf_encoder::BufEncoder;
+/// # use hex_conservative::Case;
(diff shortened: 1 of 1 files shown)


src/buf_encoder.rs
```diff
@@ -19,6 +19,22 @@ use super::{Case, Table};
...
+///
+/// # Examples
+/// ```
+/// # use hex_conservative::buf_encoder::BufEncoder;
+/// # use hex_conservative::Case;
+/// let mut encoder = BufEncoder::<4>::new(Case::Lower);
+/// encoder.put_byte(0xab);
+/// assert_eq!(encoder.as_str(), "ab");
+/// ```
+/// ```compile_fail
```
Kixunil, line 31: Documenting it is a good idea. But I'd place some explanation before the block start.

jamillambert: I added an explanation that the capacity has to be even, which does now make the the examples fit in better.

Kixunil: I meant before the failing block. Something like "The following code doesn't compile because of odd capacity".

src/buf_encoder.rs
```diff
@@ -19,6 +19,25 @@ use super::{Case, Table};
...
+/// # use hex_conservative::buf_encoder::BufEncoder;
+/// # use hex_conservative::Case;
+/// let mut encoder = BufEncoder::<3>::new(Case::Lower);  // Fails to compile due to odd capacity
+/// encoder.put_byte(0xab);
+/// assert_eq!(encoder.as_str(), "ab");
```
Kixunil, line 39: Also, if the point of the example is to show that the one line doesn't compile I'd rather replace the lines following `let mut encoder` with `# let _ = encoder;`

jamillambert: Done

src/buf_encoder.rs
jamillambert: Done

jamillambert:
Looks like this in the docs:
image

tcharding:
#155 has merged bro, you are good to rebase.

tcharding (approved):
ACK 6311325ca8c9baa534ffd1563414dfe9c6cda771

tcharding:
https://github.com/rust-bitcoin/rust-bitcoin/discussions/3337 says I'm eligible to merge stuff here. This is docs only, doesn't really benefit that much from going through @apoelstra's CI machinery.

PR #157, opened 2025-02-19
Add a test that an odd capacity `BufEncoder` does not compile

jamillambert:
An even capacity `BufEncoder` is now enforced and should be tested to make sure it does not compile if an odd capacity is used.

Adds a crate in tests that uses an odd capacity `BufEncoder`.

Adds a CI test that passes if the odd capacity crate fails to compile.

Close #155

### DIFF
--- a/contrib/compile-tests.sh
+++ b/contrib/compile-tests.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+#
+# Checks intended compile failures.
+
+set -euo pipefail
+
+REPO_DIR=$(git rev-parse --show-toplevel)
+
+compile_check() {
+    local dir_path=$1
+    local expected_status=$2
+    local result=0
+    local output
+    # All crates in the specified subdirectory must compile or fail based on expected_status
+    for dir in "$dir_path"/*; do
+        if [ -d "$dir" ]; then
+            pushd "$dir" > /dev/null
+            echo "Compiling $dir"
+            output=$(cargo build 2>&1)
+            if [ "$?" -ne "$expected_status" ]; then
+                if [ "$expected_status" -eq 0 ]; then
+                    echo "$output"
+                    echo "error: compile-tests/$dir/ failed to compile"
+                else
+                    echo "error: compile-tests/$dir/ compiled when it should not have"
+                fi
+                result=1
+            fi
+            popd > /dev/null
+        fi
+    done
+    return "$result"
+}
+
+# Check that all files in compiletests pass or fail as expected
+cd "$REPO_DIR"/tests/compile-tests
+if compile_check "pass" 0 && compile_check "fail" 101; then
+    echo "Compile tests passed"
+    exit 0
+else
+    echo "Compile tests failed"
+    exit 1
+fi
--- a/tests/compile-tests/fail/odd-capacity/Cargo.toml
+++ b/tests/compile-tests/fail/odd-capacity/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "odd-capacity"
+version = "0.1.0"
+edition = "2021"
+rust-version = "1.63.0"
+
+[workspace]
+
+[dependencies]
+hex = { package = "hex-conservative", path = "../../../../" }
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -24,3 +24,4 @@ Run from rust.yml unless stated otherwise. Unfortunately we are now exceeding th
 10. `Arch32bit`
 11. `Cross`
 12. `Format`
+13. `Compile tests`
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -218,3 +218,14 @@ jobs:
         run: rustup component add rustfmt
       - name: "Check formatting"
         run: cargo +nightly fmt --all -- --check
+
+  Compile:
+    name: Compile tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v4
+      - name: "Select toolchain"
+        uses: dtolnay/rust-toolchain@stable
+      - name: "Run compile tests"
+        run: ./contrib/compile-tests.sh
--- a/tests/compile-tests/fail/odd-capacity/src/main.rs
+++ b/tests/compile-tests/fail/odd-capacity/src/main.rs
@@ -1,0 +1,6 @@
+use hex::buf_encoder::BufEncoder;
+use hex::Case;
+
+// This should fail to compile because the capacity size is odd.
+// The only difference to `even-capacity` is the capacity size.
+fn main() { let _encoder = BufEncoder::<3>::new(Case::Lower); }
--- a/tests/compile-tests/pass/even-capacity/Cargo.toml
+++ b/tests/compile-tests/pass/even-capacity/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "even-capacity"
+version = "0.1.0"
+edition = "2021"
+rust-version = "1.63.0"
+
+[workspace]
+
+[dependencies]
+hex = { package = "hex-conservative", path = "../../../../" }
--- a/tests/compile-tests/pass/even-capacity/src/main.rs
+++ b/tests/compile-tests/pass/even-capacity/src/main.rs
@@ -1,0 +1,5 @@
+use hex::buf_encoder::BufEncoder;
+use hex::Case;
+
+// This should compile, ensuring that the compile error in odd-capacity is from the odd capacity.
+fn main() { let _encoder = BufEncoder::<4>::new(Case::Lower); }


tests/compiletests.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[test]
+fn odd_buffer_compile_fail() {
+    // Test that `BufEncoder` fails to compile when the capacity is odd.
+    let mut config = compiletest::Config::default();
+    config.mode = "compile-fail".parse().expect("Invalid mode");
+    config.src_base = PathBuf::from("tests/compiletest");
+    config.target_rustcflags = Some("-L target/debug/deps".to_string());
+    compiletest::run_tests(&config);
+}
```
Kixunil, line 13: Looks quite complicated. Why not simply have a new CI step to run `if cargo check --something; then exit 1; fi`?

jamillambert: I wasn't sure about adding a CI step just for this one, should be simple, test.  But it does work, and there might be reasons in the future to add to the compile tests?

apoelstra: I think it's probably worthwhile to do this in a general way (modifying CI). If we had the means to add compile-tests I'll bet we'd add a lot of them.

Kixunil: I prefer not to take dev dependency if the same can be done in CI.

tcharding:
Cool, looking good. I think this should be called `odd-capacity` instead of `compiletest` because we can only ever have a single test in it (since one cannot prove two build failures at the same time). And for that reason perhaps inline the function into `main`.

FTR I changed the 3 to a 4 and the crate builds - to check for no false positives. This crate risks going stale consider this:

1. We change the import path of `BufEncode` - now the test crate does not build because of that
2. We change the `BufEncode` type accidentally so that odd value builds - now the test does not catch it.

jamillambert:
I have changed it so that there are two compile checks now.  First it checks that an even capacity compiles, then it checks that the odd capacity doesn't.  The only difference between the two is changing the capacity from 4 to 3.

I also moved them to a sub directory in tests called `compiletests` so it is a bit tidier, especially if more are going to be added.

apoelstra:
Can you add a further subdirectory level `pass/` and `fail/` and have the script use the directory names to figure out whether this should be a pass or fail test?

jamillambert:
I have changed the script to be general using the two sub directories `pass/` and `fail/`.

I added the `BufEncoder` capacity evenness check as a separate commit.

tcharding:
Looks good, few `bash` things specific to the `rust-bitcoin` org

- We use `#!/usr/bin/env bash` so that Nix users can run the script (they don't typically have `/bin/bash` I'm told)
- Tend to use `set -euo pipefail` unless some reason not to
- Try to lint scripts with `shellcheck foo.sh` https://www.shellcheck.net/
- Tend to use `REPO_DIR=$(git rev-parse --show-toplevel)` to get at the repo directory
- If you feel like it (but its mainly only me that does it) you can use a bunch of helper functions that were originally written by the Rust project a long time ago. They are in various shell scripts around the place, `bitcoin/contrib/check-for-api-changes.rs` has them)

tcharding:
Can you put a `-` separator in `compiletests` directory please mate. Feel free to say 'f**k you are annoying when you review Tobin'.

contrib/compile-tests.sh
```diff
@@ -0,0 +1,57 @@
...
+
+REPO_DIR=$(git rev-parse --show-toplevel)
+
+should_pass() {
+    all_passed=true
+    # All crates in the subdirectory pass/ must compile
+    for dir in pass/*; do
+        if [ -d "$dir" ]; then
+            pushd "$dir" > /dev/null
+            if ! cargo run > /dev/null 2>&1; then
```
Kixunil, line 15: The test says "compile" but it uses `run`. Also we should keep the output here.

contrib/compile-tests.sh
```diff
@@ -0,0 +1,57 @@
...
+                echo "compile-tests/$dir/ failed to compile"
+                all_passed=false
+            fi
+            popd > /dev/null
+        fi
+    done
+    if [ "$all_passed" = true ]; then
+        return 0
+    else
+        return 1
```
Kixunil, line 25: Perhaps just define `result = 0` at the beginning and assign it to `1` on fail and just `return "$result"` to get rid of if.

contrib/compile-tests.sh
```diff
@@ -0,0 +1,57 @@
...
+    fi
+}
+
+should_fail() {
+    all_failed=true
+    # All crates in the subdirectory fail/ must fail to compile
+    for dir in fail/*; do
+        if [ -d "$dir" ]; then
+            pushd "$dir" > /dev/null
+            if cargo run > /dev/null 2>&1; then
```
Kixunil, line 35: Also it should be `build`, not `run`.

contrib/compile-tests.sh
```diff
@@ -0,0 +1,57 @@
...
+                echo "compile-tests/$dir/ compiled when it should not have"
+                all_failed=false
+            fi
+            popd > /dev/null
+        fi
+    done
+    if [ "$all_failed" = true ]; then
+        return 0
+    else
+        return 1
```
Kixunil, line 45: Same as above.

Kixunil (dismissed):

contrib/compile-tests.sh
jamillambert: done

contrib/compile-tests.sh
jamillambert: done

contrib/compile-tests.sh
jamillambert: Thanks, this is a nicer way of doing it. I have changed it as suggested.

contrib/compile-tests.sh
Kixunil: This is the `pass` branch though.

contrib/compile-tests.sh
```diff
@@ -0,0 +1,53 @@
...
+
+REPO_DIR=$(git rev-parse --show-toplevel)
+
+should_pass() {
+    result=0
+    # All crates in the subdirectory pass/ must compile
+    for dir in pass/*; do
+        if [ -d "$dir" ]; then
+            pushd "$dir" > /dev/null
+            output=$(cargo build 2>&1)
```
Kixunil, line 15: This shouldn't be needed but I guess it's OK. I don't expect the output to be too large.

contrib/compile-tests.sh
```diff
@@ -0,0 +1,53 @@
...
+
+should_fail() {
+    result=0
+    # All crates in the subdirectory fail/ must fail to compile
+    for dir in fail/*; do
+        if [ -d "$dir" ]; then
+            pushd "$dir" > /dev/null
+            output=$(cargo build 2>&1)
+            if [ $? -eq 0 ]; then
+                echo "$output"
```
Kixunil, line 35: If compilation succeeded and it shouldn't there will be no useful information printed by `cargo`. I'd rather get rid of it.

jamillambert: I have removed the cargo output for this case

jamillambert:
Looking at the script again I noticed the two functions are now basically identical.  I have refactored it to have a single function that handles both cases.

contrib/compile-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+    local result=0
+    local output
+    # All crates in the specified subdirectory must compile or fail based on expected_status
+    for dir in "$dir_path"/*; do
+        if [ -d "$dir" ]; then
+            pushd "$dir" > /dev/null
+            echo "Compiling $dir"
+            output=$(cargo build 2>&1)
+            if [ "$?" -ne "$expected_status" ]; then
+                if [ "$expected_status" -eq 0 ]; then
```
Kixunil, line 21: Why have another argument if it can be inferred from the first one and you're not comparing them anyway?

Kixunil: Too many possibilities that we're not using are just complicating the code. Honestly, I think even having the `pass` option is a stretch (normal tests do that already!) but I understand it's for the symmetry since there are related test (one for 3, other 4) so I'm accepting it. I was really expecting this to be one line in the workflow file.

jamillambert: @Kixunil you still have `requested changes` marked for this PR.  Sorry I may have misunderstood your previous comment, I read it as you are accepting it as it is?

Do you still require me to remove the expected status argument? in which case there is no reason to have a separate function at all, i.e. is below preferable the the current function in this PR?
```shell
for dir in "$REPO_DIR"/tests/compile-tests/pass/*/; do
    pushd "$dir" > /dev/null
    echo "Compiling $dir"
    if ! cargo build > /dev/null 2>&1; then
        echo "error: $dir/ failed to compile"
        exit 1
    fi
    popd > /dev/null
done

for dir in "$REPO_DIR"/tests/compile-tests/fail/*/; do
    pushd "$dir" > /dev/null
    echo "Compiling $dir"
    if cargo build > /dev/null 2>&1; [ $? -ne 101 ]; then
        echo "error: $dir/ compiled when it should not have"
        exit 1
    fi
    popd > /dev/null
done
```

Kixunil: Hmm, I think yes, I find it more readable without function than with it. I don't think this is a SSOT issue because the two loops do quite different things and only share tiny bit of code.

But I don't *require* it. I clicked "request changes" before because of run vs build confusion.

jamillambert: Thanks.  Since it's already approved I'll leave it as is, but I have the above code ready to go if needed.

contrib/compile-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+REPO_DIR=$(git rev-parse --show-toplevel)
+
+compile_check() {
+    local dir_path=$1
+    local expected_status=$2
+    local result=0
+    local output
+    # All crates in the specified subdirectory must compile or fail based on expected_status
+    for dir in "$dir_path"/*; do
+        if [ -d "$dir" ]; then
```
Kixunil, line 16: I've just realized this check turns typos in the function invocation into silent passes which is not great. I'm not going to block the PR over it but would be really nice to address.

tests/compile-tests/fail/odd-capacity/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "odd-capacity"
+version = "0.1.0"
+edition = "2021"
+rust-version = "1.63.0"
```
Kixunil, line 5: Side note: reminds me there's a proposal to have "rust scripts" - `.rs` files where `Cargo.toml` can be embedded (but not required). We could use it instead when it comes out.

Kixunil (approved):
ACK 853f493763b4a187020e1c7ed0eb7e94ef613711

PR #158, opened 2025-02-21
Remove default impl of `hex_reserve_suggestion`

tcharding:
The current default impl returns 0, this reduces performance if implementors forget to implement this method and they know up front how many bytes are needed. The trait is sealed so we are the only implementors and it turns out we forgot in one place (well many places because its a macro).

Remove the default implementation of `hex_reserve_suggestion` and implement it manually for arrays.

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -99,13 +99,11 @@ pub trait DisplayHex: Copy + sealed::IsRef + sealed::Sealed {
         })
     }
 
-    /// Hints how much bytes to reserve when creating a `String`.
+    /// Hints how many bytes to reserve when creating a `String`.
     ///
-    /// Implementors that know the number of produced bytes upfront should override this.
-    /// Defaults to 0.
-    ///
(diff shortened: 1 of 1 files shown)


src/display.rs
```diff
@@ -344,6 +342,8 @@ macro_rules! impl_array_as_hex {
                 fn as_hex(self) -> Self::Display {
                     DisplayArray::new(self)
                 }
+
+                fn hex_reserve_suggestion(self) -> usize { $len * 2 }
```
Kixunil, line 346: OMG!

Kixunil (approved):
ACK 908ce81f02719183d60b5ab151bc143c3c304654

src/display.rs
tcharding: Fail huh.

apoelstra (approved):
ACK 908ce81f02719183d60b5ab151bc143c3c304654; successfully ran local tests

PR #159, opened 2025-02-21
Enable pedantic lints

tcharding:
Configure `clippy` to warn for pedantic lints. Leave the `missing_errors_doc` one for a soon-to-come docs PR.


apoelstra:
Checked that the `let () = const` thing works the same as `let _ = const` did. There is a clippy issue saying that this case is a false positive for the lint so I wanted to be sure.

tcharding:
FTR I didn't know about `for () =` until this PR. Thanks

apoelstra (approved):
ACK 4697da858028e9b7b3f12303e7d1c0e33680d31f; successfully ran local tests

tcharding:
Bump, this guy too please.

apoelstra:
Needs rebase now.

tcharding:
Rebased and squashed. No other changes.

apoelstra (approved):
ACK a05406cdf1ae969b8bf3d7a069034e2186393586; successfully ran local tests

src/buf_encoder.rs
```diff
@@ -51,7 +51,7 @@ impl<const CAP: usize> BufEncoder<CAP> {
     #[inline]
     #[allow(clippy::let_unit_value)] // Allow the unit value of the const check
     pub fn new(case: Case) -> Self {
-        let _ = Self::_CHECK_EVEN_CAPACITY;
+        let () = Self::_CHECK_EVEN_CAPACITY;
```
Kixunil, line 54: You forgot to remove the allow

src/error.rs
```diff
@@ -79,11 +79,12 @@ impl From<Infallible> for ToBytesError {
...
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        use ToBytesError::*;
+        use ToBytesError as E;
 
         match *self {
-            InvalidChar(ref e) => write_err!(f, "invalid char, failed to create bytes from hex"; e),
-            OddLengthString(ref e) =>
+            E::InvalidChar(ref e) =>
+                write_err!(f, "invalid char, failed to create bytes from hex"; e),
+            E::OddLengthString(ref e) =>
```
Kixunil, line 87: IMO the problems the original way causes are too unlikely to bother but I'm also not going to change it back.

apoelstra: I also don't like the original way stylistically. It's hard to tell where the symbols are coming from because of the wildcard import.

Kixunil: OK, though note that for me this, as opposed to other wildcard cases, is an easily-recognizable pattern. The `match` is matching on what's clearly an enum, based on its structure and if the variants don't have the `X::` prefix then it's clear it must be wildcard imported in the statement above match. Of course one could uselessly import a different enum but that would cause other warnings. It does require a bit of discipline - to only put the `use` in the same function. But I have no problem with such discipline.

Anyway, not a big deal.

tcharding: As everyone knows I'm a sucker for uniformity. Sometimes the variant names clash and then the wildcard cannot be used which means breaking uniformity. The new style is always valid _and_ stomps the lint warning - seems like a win. It is admittedly a bit ugly.

src/buf_encoder.rs
tcharding: Fixed, thanks.

Issue #154, opened 2025-02-19
Stabilize only part of the crate


Kixunil:
For a while I felt uneasy about stabilizing the whole crate. It takes from our time to stabilize more urgent things like `units` and `primitives` and the only types we know for sure we need are the errors. Up until this point I have tolerated it thinking it would resolve easily. However I remembered I wanted `DisplayHex` to have GAT.

Our options are:

* Wait until Debian stable releases which will have GAT-capable compiler
* Bump MSRV against our policy
* Accept ugly API
* Only stabilize parts that are directly required to have stable error types

I find the first three options highly annoying, which leaves us with the last one which also has its other advantages - relieving pressure to work on hex too much.

I suggest moving everything except decoding to `hex-conservative-unstable` crate, which will reexport everything such that it'd become a drop-in replacement for the `hex-conservative` crate (people will usually only need to change name in `Cargo.toml` and it should work for most code). Then we can stabilize `hex-conservative` pretty much immediately (but practically, I'd wait until we actually have something that depends on it close to stabilization). Then shift our focus on `units` and `primitives`.

apoelstra:
Let's accept the ugly API. Adding another crate and another indefinite delay because of hypothetical features sounds awful.

If you want to use GATs we can release a 2.0, and we can semver-trick 1.1 to use the error types from 2.0, which will be functionally the same as what you're proposing, without adding a bunch of complicated indefinitely-open loops to all of our workloads.

Kixunil:
> indefinite delay

We could set a deadline to the Debian release or some offset after (1mo?)

> can semver-trick 1.1 to use the error types from 2.0

Oh, maybe yes! It needs to impl the old trait in terms of the new one but I think it's doable. I'd definitely want to try it out before committing into that approach though.

Kixunil:
Or a different possibility: have `DisplayHex` gated on compiler version. Those who need lower MSRV can just use the current `hex-conservative` for displaying. Though `bitcoin` would be one such crate (intentionally having duplicate dependencies) until the next Debian releases.

apoelstra:
Let me stew on this. It's an interesting general approach to "1.0" stuff that need a higher MSRV than our actual MSRV on 1.0-day. Or conversely, a way to artificially lower our MSRV by removing things that aren't doable with old rustcs.

In this particular case I'm a little skeptical, but I don't know the details of what you want to do with `DisplayHex` or why.

Kixunil:
[...]

PR #162, opened 2025-02-23
Prepare for initial 1.0 alpha version

tcharding:
Prepare for initial 1.0 alpha release. The target `1.x` branch will contain an initial patch that deletes everything from this repo. Then we add back in here stuff we want to release from master.

Close: #154


Kixunil:
I may have mentioned it already but I don't have much confidence in iterators. I remember when I tried to change something I ran into issues. Can we please only have the error types and two inherent functions? (I could also accept the trait.)

tcharding:
@apoelstra shall I make a `1.x` branch then set that as the merge target for this PR. Also I've still to write the changelog and then take this off draft.

src/iter.rs
```diff
@@ -194,6 +194,7 @@ where
 /// Generally you shouldn't need to refer to this or bother with it and just use
 /// [`HexToBytesIter::new`] consuming the returned value and use `HexSliceToBytesIter` if you need
 /// to refer to the iterator in your types.
+#[doc(hidden)]
```
Kixunil, line 197: Why is this hidden and not non-pub/deleted?

apoelstra:
> @apoelstra shall I make a `1.x` branch then set that as the merge target for this PR?

Yep.

tcharding:
After sleeping on this the following thought came to me:

If we are using `doc(hidden)` to hide unstable stuff why not just hide everything in 1.0 we don't want to commit to and not bother with doing 0.4 at all? Then there is only one dependency and we can break what we think we might have to - YOLO.

apoelstra:
@tcharding because if anybody (including us) actually *uses* the `doc(hidden)` stuff then we still cant change their API without breaking downstream users' builds.

tcharding:
Open question: I have no idea how we will re-introduce stuff back into this from another branch and have reviewable patches.

tcharding:
`1.x` branch created and merge target set.

Kixunil:
> Open question: I have no idea how we will re-introduce stuff back into this from another branch and have reviewable patches?

Any 1.0 -> 0.x should only happen by adding `use`. Any 0.x -> 1.0 should only happen by code moves. I think it'll be fine.

apoelstra:
> Any 0.x -> 1.0 should only happen by code moves.

It might look a little weird in git. Normally you don't "move" code between branches like this.

> I think it'll be fine.

But yeah, agreed. We're all pretty good at using git here and we'll figure it out.

Kixunil:
> It might look a little weird in git.

Oh, true, one commit deletes it another adds it. We will need to link them somehow at least.

tcharding:
After talking with @apoelstra today I pushed a few changes

`1.x`:
-  Has had all the history removed (reset to the initial commit)
- Has two patches on it. One introduces non-code stuff from `master` and the next introduces an empty crate and CI.

This branch:
- Has a wip (waiting for PRs to merge) patch that introduces the code we want to release
- Has a patch bumping the version number

Please concept ack/nack this. Another solution would be to put the two patches on `1.x` at the front of this PR

tcharding:
Tested in https://github.com/rust-bitcoin/rust-bitcoin/pull/4249

apoelstra:
>Honestly, I'm not sure if I really want the renames, I just want the other names to be considered

The existing name is so weird that I'd prefer to change it. It would be fine if it were just ambiguous, or just too short or too long, or whatever. But I can't think of any other error that starts with `To`.

> * `DecodeHexToArrayError` - this is clear and has better English word ordering (which is recommended in API guidelines) but it's quite long. Though I don't expect the type to be typed too often. `rg` says there are currently 11 occurrences in the entire `rust-bitcoin` repository.

This is my preference. We will very rarely type this name. Better to be unambiguous.

tcharding:
> Change the names back to Hex* - I guess it's not really needed because of above

FWIW the names were not changed. The `HexToArrayError` type still exists, the `ToArrayError` is an inner error that is "just the variants we have right now and we want users to be able to get access to in a way that we can still change the outer error" - I did not know a name for this so I just made the inner error name shorter that the outer one. At the time I did not think of `hex::ToArrayError`.

tcharding:
if its useful here is an example usage of this error. https://github.com/rust-bitcoin/rust-bitcoin/pull/4249/commits/6b9fef2e959f29ce55fd5a38bdcd97437af862f5#r2000056324

Kixunil:
Ah, but if they are inner errors why are they public?

For private inner errors I usually use `FooErrorInner` in my code.

Sometimes these things are named `ErrorKind` but they don't usually carry additional data (I wouldn't mind though). If we want to make the enum public, we should just change the outer type to enum. This will have a cost of branch if we at some point add a field to both variants that is frequently accessed regardless of the kind (the only thing that I can imagine here is the input string) but I think that's fine and I believe the cost can be hacked out by manually managing the layout of the types stored in variants such that the common fileds land on same offset - the optimizer will then see `if discriminant { ptr.offset(x) } else { ptr.offset(x) }` and remove the condition.

Now that I typed it I think just exposing the enum is the most reasonable approach. If you disagree, feel free to close it completely.

In other words, the library should have **exactly two** error types. (For now.)

tcharding:
Leaving for @apoelstra to answer.

apoelstra:
These changes were made in #129.

The "inner" errors are public because downstream users want to match on the existing variants, but they're still wrapped because this gives us a bunch of extra API freedom. See https://github.com/rust-bitcoin/hex-conservative/pull/129#issuecomment-2565966346 -- the idea is that we can completely rewrite `HexToBytesError` as long as there's some way that we can still produce a `ToBytesError` from it.

I don't know why they're named the way they are. In that thread I had suggested a `Inner` suffix, and then accepted whatever Tobin suggested because I didn't see any reason to care about an auxiliary type name.

@Kixunil if you think we can just expose the enum as the top-level error then let's do it. But this whole exercise was attempting to API-proof stuff in the hopes that you wouldn't hold up 1.0 forever about it.

Kixunil:
Hmm, actually I've realized that if we ever need to do feature-based boxing then having them closed has better outcome. OK, let's keep them closed.

But in that case, I want the inner errors to be named `*Kind` (yes, carrying additional data is unusual but I don't mind) and rename the accessor function to `first_error_kind` or just `first_error`. (Noting that if we have multiple errors in the future, this returns the kind of first one.)

But regarding the code that's using them - it uses them to panic on invalid lengths. I think it should just be rewritten to not panic by attempting to parse both lengths - similar to what I did in another PR with `if let Ok(array) = slice.try_into() { } else if let Ok (array) = slice.try_into() { } else { Err(...) }`. I had a bunch of other ideas around this but they are too complicated to implement today (maybe in a few years they'll be feasible).

tcharding:
Leaving the rename to you @Kixunil, I'm setting this as draft till its done.

Kixunil:
Should I send PR against the 1.x branch?

tcharding:
Against master please, I'll handle update this one after that merges. Cheers man

tcharding:
Holy f*** have a look at this.

- `FromHex` associated error is private
- New error types, totally public, for the new decode functions

This gives us total freedom going forward to modify `FromHex`. The only guarantee we make is that we will keep the two decoding functions and they will return the two decoding errors - BOOM!

tcharding:
Review, ack, merge, release, high five, drink, dance naked round the fire.

src/iter.rs
```diff
@@ -0,0 +1,414 @@
...
+    }
+}
+
+/// An internal iterator returning hex digits from a string.
+///
+/// Generally you shouldn't need to refer to this or bother with it and just use
+/// [`HexToBytesIter::new`] consuming the returned value and use `HexSliceToBytesIter` if you need
+/// to refer to the iterator in your types.
+#[derive(Debug)]
+pub struct HexDigitsIter<'a> {
```
Kixunil, line 196: This was not supposed to be public and I *think* it's not publicly accessible but the `pub` makes it hard to check. Could you please change it to `pub(crate)`?

src/error.rs
```diff
@@ -0,0 +1,219 @@
...
+pub(crate) use write_err;
+
+/// Hex decoding error while parsing to a vector of bytes.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DecodeToBytesError {
+    /// Non-hexadecimal character.
+    InvalidChar(InvalidCharError),
+    /// Purported hex string had odd length.
+    OddLengthString(OddLengthStringError),
+}
```
Kixunil, line 37: FTR, I had a reason to actually not open this. But now I've realized it'd be much nicer if we could get the optimization I'm thinking of into the compiler and I believe it's possible. So ACK this layout.

Kixunil (approved):
ACK 8b6ee9c7528c61d28277390916694e1be379c9cf

src/iter.rs
tcharding: Done. Also I grepped for 'pub ' and made anything else that is not in the public API `pub(crate)` for clarity. Now the output of grep is
```bash
gg 'pub '
src/error.rs:32: pub enum DecodeToBytesError {
src/error.rs:81: pub enum DecodeToArrayError {
src/error.rs:128:pub struct InvalidCharError {
src/error.rs:141:    pub fn invalid_char(&self) -> u8 { self.invalid }
src/error.rs:144:    pub fn pos(&self) -> usize { self.pos }
src/error.rs:159:pub struct OddLengthStringError {
src/error.rs:171:    pub fn length(&self) -> usize { self.len }
src/error.rs:186:pub struct InvalidLengthError {
src/error.rs:201:    pub fn expected_length(&self) -> usize { self.expected }
src/error.rs:204:    pub fn invalid_length(&self) -> usize { self.invalid }
src/lib.rs:24: pub mod _export {
src/lib.rs:26:     pub mod _core {
src/lib.rs:27:         pub use core::*;
src/lib.rs:42: pub use self::{
src/lib.rs:55: pub fn decode_vec(s: &str) -> Result<Vec<u8>, DecodeToBytesError> {
src/lib.rs:73: pub fn decode_array<const N: usize>(s: &str) -> Result<[u8; N], DecodeToArrayError> {
src/parse.rs:48:     pub trait Sealed {}
```

Kixunil: Cool!

We need to hide `invalid_char` for now since it'd be better to return `char` instead but we didn't implement it yet.
We can also completely remove `_export` because we don't have any public macros.
And we can also remove `Sealed` because there's no public trait to be sealed.

tcharding: - Removed `__export`
- Removed `sealed` module entirely
- Hid `invalid_char` (FTR I don't agree with this but its done).

src/error.rs
```diff
@@ -0,0 +1,222 @@
...
+                write!($writer, concat!($string, ": {}") $(, $args)*, $source)
+            }
+        }
+    }
+}
+pub(crate) use write_err;
+
+/// Hex decoding error while parsing to a vector of bytes.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DecodeToBytesError {
```
Kixunil, line 32: Sorry for poking into naming again but I just realized this name is inconsistent with `DecodeToArrayError` since this one doesn't refer to a type but `Array` does. And I also realized mentioning the type leaks internal details about functions that call our functions.

So I think we should refer to the property that distinguishes them: length being fixed or dynamic. Thus `DecodeDynSizedError` and `DecodeFixedSizedError`.

But if you're really annoyed by this bikeshedding, I can accept "shut up and get it released". I kinda suspect the downstream errors will want to map the enum to something else instead so "leaking internals" is not going to happen anyway.

src/parse.rs
```diff
@@ -0,0 +1,218 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Hex encoding and decoding.
```
Kixunil, line 3: No encoding, just decoding. Also I'd rename the module to `decode`. Feel free to leave to another PR.

src/lib.rs
```diff
@@ -7,7 +7,8 @@
 //! ## Stabalization strategy
```
Kixunil, line 7: Typo: s/Stabalization/Stabilization/

@tcharding I noticed you're using this word all over the place. So far I just assumed it's Australian dialect. However since it appeared in the public doc I started digging into it and it seems it's not really that and just misspelling. (Can't duckduckgo the word and ChatGPT also says it's a mistake.)

Feel free to explain but I'm pretty confident that "stabilization" is the correct US English term and we've decided somewhere else that our public communication will be US English.

src/lib.rs
```diff
@@ -7,7 +7,8 @@
...
 //!
 //! In an effort to release stable 1.0 crates that are forward compatible we are striving
-//! relentlessly to release the bare minimum amount of code.
+//! relentlessly to release the bare minimum amount of code. This 1.0 version currently holds
+//! only two hex parsing functions and the associated error types.
```
Kixunil, line 11: I'd like this doc to be expanded before release to mention that stable crates should use the error types from this library in their public APIs but can use the 0.x.y version so far for advanced features. If you don't feel like inventing wording I can send a PR later.

src/lib.rs
```diff
@@ -19,10 +20,57 @@
...
+
+use crate::parse::FromHex as _;
+
+#[rustfmt::skip]                // Keep public re-exports separate.
+#[doc(inline)]
+pub use self::{
+    error::{
+        DecodeToArrayError, DecodeToBytesError, InvalidCharError,
+        InvalidLengthError, OddLengthStringError, 
+    },
```
Kixunil, line 38: FTR I think these re-exports are alright. I expect `hex::WhateverError` to be common and I like it.

src/lib.rs
```diff
@@ -19,10 +20,57 @@
...
+};
+
+/// Decodes a hex string into a vector of bytes.
+///
+/// # Errors
+///
+/// Errors if `s` is not a valid hex string.
+#[cfg(feature = "alloc")]
+pub fn decode_vec(s: &str) -> Result<Vec<u8>, DecodeToBytesError> {
```
Kixunil, line 47: Despite myself mentioning not referring to target types in error types, referring to them in **functions** is correct and appropriate because that's the specific type that the function returns.

However, because "to decode x" means that "x" is source, not the target, as we discussed already, this should be `decode_to_vec`.

src/lib.rs
```diff
@@ -19,10 +20,57 @@
...
+        }
+    }
+}
+
+/// Decodes a hex string into an array of bytes.
+///
+/// # Errors
+///
+/// Errors if `s` is not a valid hex string or the correct length.
+pub fn decode_array<const N: usize>(s: &str) -> Result<[u8; N], DecodeToArrayError> {
```
Kixunil, line 65: As above, this should be `decode_to_array`.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 1.0.0-alpha.0 - 2025-02-27
```
Kixunil, line 1: This is extremely close to release IMO. I wouldn't be even annoyed if there was push to skip the pre-release. (To be clear, yes, let's do a pre-release just to be absolutely sure.) So I think `rc` would be more accurate name - this definitely is a release candidate in my mind.

But it's not super important to me, just something to reduce confusion.

Kixunil (commented):
I think this is the final round of bikeshedding.

tcharding:
We are back at 'do it yourself' @Kixunil - remember last time I went off my nut at you I said that I did not think you would ever sign off on a stable release? Can you please PR and get this fucking crate in a state that you are happy with. And can you do it within two weeks please.  FUCK.

src/lib.rs
tcharding: If its not obvious to you why I'm so pissed (again) - you suggested these fucking names in the first place.

apoelstra:
@tcharding in the interest of absorbing some of your frustration, I also knocked my local CI box out for the last 36 hours (moving my nix store from a HDD to a SSD) and it'll probably be out, or at least degraded, for another 24 hours while I rebuild some supporting derivations that were deleted during the migration (mainly, a ludicrously long-running rocq proof of some sort in Simplicity).

Though for code review I should be able to  mostly move to my "slow" box.

src/lib.rs
tcharding: Fixed! I now have the `hex-conservative` and `rust-bitcoin` directories configured to use US spelling. Could have done that two years ago.

src/lib.rs
Kixunil: Sadly, I'm just a human.

Kixunil:
@tcharding seems like you're volunteering to do things that get delayed too often. I suggest we try to swap roles, since I'm fine with delays and discussions of these fundamental API things. I'm myself not 100% sure of all things I say that's why I often ask. Notice this didn't get "request changes" from me because if we released it as-is it wouldn't be *terrible* but it seems that we can do better with reasonably small effort.

Anyway, I'll use this PR as a template and do the changes that I think make sense since my comments here didn't get any specific counterarguments, so I assume you just don't feel like implementing them rather than thinking they're wrong.

As a personal recommendation, I recommend looking into Rust's RFC discussion. Many of them. Perhaps starting with features you want in Rust. They gave me important perspective on API design and that's where my careful approach comes from. It happened to me so many times that something seems "obviously right" until someone came up with an argument that undermines my conviction. Doing this stuff is really hard.

src/lib.rs
apoelstra: >Sadly, I'm just a human.

Kix, this is fine, but you personally have been holding up multiple crates in this ecosystem, for various combinations of things that don't matter (like this, whether you are "decoding a hex-encoded array" or "decoding a hex string to an array", what the exact value of MAX_MONEY should be), things that require months of work from you to work around problems inherent to the language (e.g. everything to do with `io`, the `Address<Unchecked>` mess which manages to be super annoying while not actually representing that the address' network matches any particular value, the mess with `PushBytes` which is similarly super annoying while not keeping scripts within any length limits, the vaporware "push decoding" project, etc).

It turns everything into a never-ending argument and it's next to impossible to get anything because you are the only one who writes code that you'll let pass. (Recall how long you complained about my `From<u32> for Sequence` which I eventually just gave up on, and then you went ahead and put a bunch of `From<&[&[u8; N]]>` impls on `Witness` recently, which neither Tobin or I complained about because obviously you should be able to convert a weakly-typed object to a strongly-typed version. How was that a good use of any of our time to argue about `From<u32> for Sequence` for weeks and then have no change to the library except that it's harder for me to use for no reason??)

The result of this arguing is bad code by the way -- Tobin and I spend all our energy going in circles about API design and then we have less time and energy to carefully review code, something which is **way harder** because there are always multiple never-to-be-merged projects happening in parallel and we know that most of them won't ever be merged so what's the point.

I appreciate that you acknowledge your fallibility. But then please stop turning everything into drawn-out fight. If there are things you really care about then prioritize those. Because I am going to start overriding your complaints without thinking about them carefully because there's just too many of them and they're preventing any of the good things you bring to the project from seeing the light of day.

apoelstra:
> As a personal recommendation, I recommend looking into Rust's RFC discussion. Many of them. Perhaps starting with features you want in Rust. They gave me important perspective on API design and that's where my careful approach comes from. 

You see "no specialization at all for over a decade, and not for another decade at least" and want to imitate that?

Kixunil:
> no specialization at all for over a decade, and not for another decade at least

And do you know the reason for that?

apoelstra:
>And do you know the reason for that?

People being unwilling to accept that there is no perfect solution and it would be better to ship *something* than nothing, and instead cheating by enabling specialization for `std` in ways that aren't available to the rest of us.

(And yes, I am aware of specific soundness issues, which is why for example `Clone::clone` can specialize if `Copy` is implemented, but even in std, nothing can specialize to a trait that has methods you can implement. In effect, even std only gets specialization of the form "you can detect whether a trait is implemented and branch based on that fact, but you cannot actually use the trait". It doesn't change my broad point, which is that it is possible to specialize soundly in many cases, and the Rust developers have instead chosen to allow zero cases.)

Kixunil:
I was under the impression that the soundness issues are deeper and the only way `std` gets around them is by being very careful with their code (basically `unsafe`). But if it's true that specialization on empty traits would've been sound then it's really annoying it's not available. However, what would already help, and it was so far rejected, is the ability to get type ID for non-`'static` types (with a big fat warning that if you use it to cast types that have lifetime parameters all hell is going to break lose).

apoelstra:
> I was under the impression that the soundness issues are deeper and the only way `std` gets around them is by being very careful with their code (basically `unsafe`).

I *think* they tried this and were still wrong, so they then just ripped out everything except for "specializing on marker traits". But I'm unsure. I discovered this after-the-fact and then I haven't thought about it in several months, since the last time I tried to improve rust-miniscript's representations and had to deal with this.

Kixunil:
Well, sounds like that proves my point: it's bloody hard.

But there are hacks to get specialization, just today I've learned about two - exploit the observable effect of `std` specializing on `Copy` (if you make `Clone` write to a `Cell` then specialized impls will cause that write to not execute and you can detect that), so you can get marker trait specialization and `transmute` of `Trait` to `Trait + 'static` being sound (because lifetimes don't affect layout), so that you can get `TypeId` without `'sized` bound. Pretty cool stuff. The only thing that's missing is stuff like calling `fill_buf` on a reader that implements `BufRead`.

PR #163, opened 2025-02-24
WIP: Depend on hex 1.0.0-alpha.0

tcharding:
We are attempting to do some release trickery in order to stabalize hex more quickly. It was observed that we only need the `hex` error types in the public API of downstream bitcoin crates that we want to stabalize. So we created a 1.0 that was just the parsing logci and error types.

Depend on the new `hex 1.0.0-alpha.0` crate and re-export all its types and traits in the same spot as they were here.


tcharding:
This tests #162. Expect CI to fail because we patch the manifest.

tcharding:
Ugh, we still can't do this with the make-everything-private version in #162 because we cannot remove `HexToBytesIter` (since its private in 1.0) and we cannot keep it because it requires error constructors.

How much developer time are we willing to waste chasing this unicorn of the perfect 1.0 crate?

tcharding:
I believe this is a lost cause.

PR #164, opened 2025-02-24
Add tests to kill all mutants in `hex-conservative`

jamillambert:
All the mutants found by `cargo mutants` have been killed by adding tests or when needed adding an exception to `mutants.toml`

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,0 +1,12 @@
+additional_cargo_args = ["--all-features"]
+examine_globs = ["src/**/*.rs"]
+exclude_globs = [
+]
+exclude_re = [
+    "serde", # Skip serde mutation tests
+    "Iterator", # Mutating operations in an iterator can result in an infinite loop
+    "write_pad_left", # Timeout
+    "write_pad_right", # Mutant changes `* 1` to `/ 1` which has no effect
+    "hex_reserve_suggestion", # Mutant changes the suggested number bytes to reserve when creating a `String` to 0 or 1. This has a possible performance impact but is not a correctness issue.
+    "fmt_hex_exact_fn", # Mutant changes a `/ 2` to `* 2` in the number of elements in the iterator. Only the required number of elements are returned so the mutant has no functional change to test.
+]
--- a/src/buf_encoder.rs
+++ b/src/buf_encoder.rs
@@ -252,6 +252,19 @@ mod tests {
     }
 
     #[test]
+    fn put_filler() {
+        let mut encoder = BufEncoder::<8>::new(Case::Lower);
+        assert_eq!(encoder.put_filler(' ', 0), 0);
+        assert_eq!(encoder.as_str(), "");
+        assert_eq!(encoder.put_filler('a', 1), 1);
+        assert_eq!(encoder.as_str(), "a");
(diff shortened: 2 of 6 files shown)


src/buf_encoder.rs
```diff
@@ -251,6 +251,19 @@ mod tests {
...
+    fn put_filler() {
+        let mut encoder = BufEncoder::<8>::new(Case::Lower);
+        assert_eq!(encoder.put_filler(' ', 0), 0);
+        assert_eq!(encoder.as_str(), "");
+        assert_eq!(encoder.put_filler('a', 1), 1);
+        assert_eq!(encoder.as_str(), "a");
+        assert_eq!(encoder.put_filler('é', 2), 2); // Test 2 byte UTF-8
+        assert_eq!(encoder.as_str(), "aéé");
+        assert_eq!(encoder.put_filler('é', 4), 1); // Try to fill more than fits
+        assert_eq!(encoder.as_str(), "aééé");
```
Kixunil, line 264: Oh, this made me realize our test that `BufEncoder` has even length is likely useless because someone might need to fill in odd number of bytes. :facepalm:

.cargo/mutants.toml
```diff
@@ -0,0 +1,12 @@
...
+examine_globs = ["src/**/*.rs"]
+exclude_globs = [
+]
+exclude_re = [
+    "serde", # Skip serde mutation tests
+    "Iterator", # Mutating operations in an iterator can result in an infinite loop
+    "write_pad_left", # Timeout
+    "write_pad_right", # Mutant changes `* 1` to `/ 1` which has no effect
+    "hex_reserve_suggestion", # Mutant changes the suggested number bytes to reserve when creating a `String` to 0 or 1. This has a possible performance impact but is not a correctness issue.
+    "fmt_hex_exact_fn", # Mutant changes a `/ 2` to `* 2` in the number of elements in the iterator. Only the required number of elements are returned so the mutant has no functional change to test.
```
Kixunil, line 11: This is confusing to me, which line is it?

src/buf_encoder.rs
jamillambert: I wasn't sure why in the code `buf_encoder.rs:136` it has `let max_capacity = self.buf.remaining_capacity() / filler.len();` when for hex `filler` (char) always has a length 1.  I assumed there must be use case for non hex UTF-8 characters, or at least the code was made to work with them.

jamillambert: The `BufEncoder` capacity in characters must be even, not the size in bytes.  I though you were talking about the two byte non hex character in the test at first.

tcharding (approved):
ACK ca2319470e626f696fcfe1a6503280103117846e

tcharding:
Typically we try not to test against the string output of errors because its a maintenance burden but here in hex because we are trying to 1.0 the errors I think it may add value. I believe though we agreed to not consider debug output of errors as stable. cc @Kixunil and @apoelstra

Kixunil:
Yes, `Debug` is never stable, `Display` is only stable on some types, generally all error types are not stable.

jamillambert:
I have removed all checks against error strings.  Checking the error type and that it has a string is enough to kill the mutants.

tcharding (approved):
ACK 0aa5ab54cdd55ecc0061957827031d020923a696

PR #166, opened 2025-03-03
Introduce crate level decoding functions

tcharding:
We would like to release a bare-bones 1.0 version of this crate that excludes the `FromHex` trait. So that users can still decode vectors and arrays without the trait introduced two new decoding functions.

Intentionally do not use `decode` as the name incase we later want to add a more general `decode` function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,6 +66,9 @@ pub mod prelude {
     pub use crate::{display::DisplayHex, parse::FromHex};
 }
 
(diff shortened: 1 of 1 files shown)


tcharding:
Pulled out of #162

apoelstra (approved):
ACK 97222f253d4ec6415192c11c7ec6bbddeda27a59; successfully ran local tests

PR #172, opened 2025-03-25
Release 1.0-rc.0

Kixunil:
Alternative to #162 with a bunch more cleanups and improvements.

I'm fairly confident this is as close to releasable as I can get, however please do check the API and please do complain if you see a problem.

If you see problems unrelated to the API please say so but we can fix them between 1.0.0-rc.0 and 1.0.0 so I'll only apply such fixes if there are no ACKs and I have enough time.


apoelstra:
Are these conflicts real or a Github hiccup? Can you rebase on master?

apoelstra:
Err, rebase on 1.x?

IIRC Tobin did a force-push to the 1.x branch which somehow changed the starting point.

Kixunil:
Oh, yeah, OK.

src/lib.rs
```diff
@@ -1,20 +1,152 @@
...
+///
+/// The length of the returned `Vec` is determined by the length of the input, meaning all even
+/// lengths of the input string are allowed. If you know the required length at compile time using
+/// [`decode_fixed_sized`] is most likely a better choice.
+///
+/// # Errors
+///
+/// Errors if `hex` contains invalid characters or doesn't have even length.
+#[cfg(feature = "alloc")]
+pub fn decode_dyn_sized(hex: &str) -> Result<Vec<u8>, DecodeDynSizedBytesError> {
```
apoelstra, line 79: In 040a20e1787922dbe483800406ad3c81ef07de82:

Both `dyn` and `sized` mean things in Rust that have nothing to do with this funtion. Calling this `decode_vec` was fine. You could also call it `decode` because the obvious thing that a hex string decodes to is a byte vector.

src/lib.rs
```diff
@@ -1,20 +1,152 @@
...
+
+/// Decodes a hex string with expected length kown at compile time.
+///
+/// If you don't know the required length at compile time you need to use [`decode_dyn_sized`]
+/// instead.
+///
+/// # Errors
+///
+/// Errors if `hex` contains invalid characters or has incorrect length. (Should be `N * 2`.)
+pub fn decode_fixed_sized<const N: usize>(hex: &str) -> Result<[u8; N], DecodeFixedSizedBytesError> {
```
apoelstra, line 91: In 040a20e1787922dbe483800406ad3c81ef07de82:

`sized` means something in Rust that has nothing to do with this function. I guess you mean `fixed_size`, but then this is in contradiction with the commit message where you say that maybe we'll also want a method to decode fixed-size objects to `Vec<u8>`.

`decode_array` was fine. `decode_fixed_size` would also be ok, even though it's longer and less descriptive and might conflict with future functions.

src/error.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// If you're parsing a larger string that consists of multiple hex sub-strings and want to
+    /// return `InvalidCharError` you may need to use this function so that the callers of your
+    /// parsing function can tell the exact position where decoding failed relative to the start of
+    /// the string passed into your parsing function.
+    ///
+    /// Note that this function has the standard Rust overflow behavior because you should only
+    /// ever pass in the position of the parsed hex string relative to the start of the entire
+    /// input. In that case overflow is impossible.
+    ///
+    /// This method is specifically designed to be used with [`map_err`](Result::map_err) method of
```
apoelstra, line 182: In ba658f2af371fe92bcc393e2eb904931a9b3e11f:

This line feels weird. You'd think if it were specifically designed for `Result::map_err` then it'd match the signature of the argument `Result::map_err`, which it can't because it takes extra data.

Better to just drop this line and trust that users are familiar with the Rust standard library.

Kixunil: The point of the line is that this is why it returns `Self` rather than taking `&mut`, but maybe we should go all the way:

> which it can't because it takes extra data.

There's an idiom I tend use sometimes:

```rust
fn add_error_context(context: Context) -> impl FnOnce(SomeError) -> MaybeDifferentError {
    move |error| {
    // do the conversion here
    }
}
```

Which actually can be used as an argument just fine: `.map_err(add_error_context(context))`. This is obviously not a good idea with contexts that are expensive to construct but fine with integers.

But because it's unusual (I have literally never seen anyone else doing this) that seems too much to me.

BTW, since the input is consumed, the current design makes modifying `&mut` quite annoying. But I think these kinds of mappings will be the only place where people will reasonably use this.

src/lib.rs
Kixunil: Yeah, I'm not happy with that either. But is there a better way in English to express "this doesn't have fixed length"?

src/lib.rs
Kixunil: I definitely didn't mean `size` since that'd sound like we're decoding some kind of size.

Also it's not really that far off Rut terminology - arrays are `Sized` and whatever is backing a `Vec` is not.
There is no contradiction - if a function is needed we can add it later. If `Box::write` was stable I might've just ignored this and tell people to use it with this function since it should optimize well.

src/error.rs
apoelstra: Heh, yeah, that did occur to me after I'd posted my comment. But agreed, it's too unusual and almost looks like a mistake when it's used, so better to just stick with the common (if ugly) thing.

src/lib.rs
apoelstra: If we use `fixed_size` it sounds like we're decoding a fixed-size object. It does not sound like we're decoding a `size`.

And `Vec`s are `Sized`. This terminology is confusing.

src/lib.rs
apoelstra: A `&str` does not have a fixed length. Nor does a `Vec<u8>`. We don't need to express this. It's in the type signature, and the type signature is the most natural thing that any function called `hex::decode` could have.

If you don't like `decode_vec` I think we should go with `decode`.

Kixunil: I'd rather go with `_vec` suffix to not make vecs more convenient than arrays.

Kixunil:
I might revert the function names to `_vec` and `_array` but I'm going to take a bit of time to think about it.

apoelstra:
@Kixunil have you given any thought to this? I am tempted to just move forward with #162 instead.

tcharding:
Why is all the new stuff not being pushed to master first?

tcharding:
If I am understanding patch 1 (ba658f2) it conceptually does a few things:

- Removes wrapper error type and just leaves the 'inner' errors
- Improves the inner errors by adding `offset` function and renames them
- Improves the naming of top level decode functions

So I don't see why this all is not PR'd against master then this PR becomes basically a simple copy paste of files on master.

Kixunil:
@tcharding oh, I've misunderstood the process. I was just going to do the renames now but seeing that I'll look into making a PR to master.

tcharding (approved):
ACK 4b15b7f76527d2a60f536ba43636102b263bbaf3

tcharding:
I've pushed two patches to the `1.x` branch. One to remove the lock file and another to add gitignore.

The commit log in ba658f2af371fe92bcc393e2eb904931a9b3e11f is stale, needs a bunch of stuff removing since we merged into master.

Kixunil:
I'm kinda confused what should the next steps be. The PR merged into `master` is not a perfect clone of this - intentionally. We need slightly different docs in different branches/releases.

I'm thinking I should mostly just copy the files and apply relevant patches but maybe you had a different process in mind?

tcharding:
IMO this is ready for release merge and release. @apoelstra can you cut the release please?

Kixunil:
Looking at the diff between `master` and this branch:

* The feature comments in `Cargo.toml` differ (NBD IMO)
* There's redundant `documentation` key
* Different dependencies - intentional
* Removed lints - why?
* Different intro in README - intentional
* Different style of MSRV - a bug IMO, `master` has nicer one but fine to release as-is in rc and fix it up later
* Bunch of random doc changes - not urgent but I would love to clean them up at least in 1.0

So yes, I'm technically fine with the rc, though I will try to clean it up if possible. I plan to open a new PR so that either it gets reviewed in time and can replace this one or it doesn't and this one can go in.

apoelstra:
Hiya. Sorry for ignoring this. My parents were over last week and I built up a backlog I had to get through. Lemme review this and let's go.

apoelstra:
Ok, NACK the name `Sized`. At **best** this is ungrammatical and weird-sounding. But as I have repeatedly argued, it is also misleading and wrong.

Kixunil:
> ungrammatical

Oh, interesting, I though statically-sized is proper English.

apoelstra:
"Statically-sized" is okay. "Fixed-sized" is okay if you interpret "fixed" as an adverb modifying the adjective "sized" but it's not the natural way to read it, because

* We have `FixedSizedError` where `Sized` clearly does not refer to the error (I think) so it has no referent and sounds it should be the noun
* "fixed size" is a very common phrase, while "fixed sized" is essentially never used, which further leads the reader to read "fixed" as a *adjective* modifying the *noun* "size"

apoelstra:
@Kixunil can you remake this PR by just taking `master` and deleting all the stuff you don't want in the rc?

Kixunil:
Yeah, I will try to look into it tomorrow.

apoelstra:
I am going to do the followups to https://github.com/rust-bitcoin/rust-secp256k1/pull/716 and then I'm going to take this on. So today or tomorrow.

(And yes, I know that I'm no more likely than you to actually hit a "tomorrow" deadline :) but let's get this thing done!)

tcharding:
LFG

apoelstra:
Closing in favor of https://github.com/rust-bitcoin/hex-conservative/pull/175

PR #173, opened 2025-04-04
Cleanups to prepare 1.0 release

Kixunil:
This includes the relevant stuff from #172 along with the changes requested in the review.


README.md
```diff
@@ -1,14 +1,47 @@
...
 using the MSRV toolchain.
 
-### Githooks
+### Policy
+
+We don't intend to bump MSRV until the newer Rust version is at least two years old and also
+included in Debian stable (1.63 is in Debian 12 at the moment).
+
+Note though that the dependencies may have looser policy. This is not considered breaking/wrong
+- you would just need to pin them in `Cargo.lock` (not `.toml`).
```
tcharding, line 41: This renders as a list because the `-` is at the start of the line

Seen here: https://github.com/Kixunil/hex-conservative/blob/1.0-cleanups/README.md

src/error.rs
```diff
@@ -13,101 +48,101 @@ use core::fmt;
...
+pub(crate) use write_err;
 
-/// Hex decoding error.
+/// Error returned when hex decoding a hex string with variable length.
+///
+/// This represents the first error encountered during decoding, however we may add other remaining
+/// ones in the future.
+///
+/// This error differs from `DecodeFixedSizedBytesError` in that the number of bytes is only known at run time
+/// - e.g. when decoding `Vec<u8>`.
```
tcharding, line 74: This renders as a list item because of the `-` at the start of the line.

tcharding: We could use a link for `DecodeFixedSizedBytesErro` after 'differs from' since reader is likely to want to see the `DecodeFixedSizedBytesError` when reading.

tcharding: This fixes both things

```rust
/// This error differs from [`DecodeDynSizedBytesError`] in that the number of bytes is known at
/// compile time - e.g. when decoding to an array of bytes.
```

tcharding (commented):
This is a nice PR, appreciate the effort you went to. A couple minor docs things only. I did not review the grammar of docs too closely. I think we can do a final parse of docs during the rc cycle.

tcharding:
With the docs fixes:

ACK 579bdff0a052fbecb4de450139e014690cbef36c

README.md
Kixunil: LOL! Nice catch

Kixunil:
I fixed the docs as suggested.

tcharding (approved):
ACK d73c6866e721f33cf836ca3d31df48cc8b10421a

apoelstra:
Nice! I'm happy with these names. Am in flight today but will review this now so that it can get into the queue before I'm offline.

Disappointing that copilot is still useless.

apoelstra:
Will ACK and merge this regardless, but I'd like a followup to

* rename the `newer-rust-version` feature to `rust-1..81` to make it obvious what "newer" means, which is currently meaningless and will only become more meaningless
* update the README text which I think is overly targeted at developers of this library
* Fix the error type names which have `Sized` in them but which have nothing to do with `Sized`

I can do this. It will be a much smaller PR than this one.

apoelstra (approved):
ACK d73c6866e721f33cf836ca3d31df48cc8b10421a; successfully ran local tests

Kixunil:
> Disappointing that copilot is still useless.

It did find two issues without me having to do anything other than clicking. I consider it a win. Maybe it's value is limited but still, I didn't have to wait for you to find those mistakes and fixed them quickly.

> * rename the `newer-rust-version` feature to `rust-1..81` to make it obvious what "newer" means, which is currently meaningless and will only become more meaningless

I disagree with putting specific version in the feature name. *Currently* it's only about `core::error::Error` but we might want to use some other feature as well and then the version would be just wrong (or we would need multiple features which then begs the question why have the dependency at all). The real end version is the one which will stabilize `#[cfg(lang-version = "...")]` one day in the future and we currently don't know which it is.

> * Fix the error type names which have `Sized` in them but which have nothing to do with `Sized`, and have `Dyn` in them but have nothing to do with `dyn`.

As I explained elsewhere, it is very similar to `Sized`. But the only other name that comes to my mind is `VarLen` which is also a bit confusing. Originally I put the rename as the last commit so that it could be trivially dropped from this PR.

apoelstra:
I strongly disagree that there is any similarity between `Sized` and the difference between `[u8; N]` and `Vec<u8>` (which are both `Sized` types), or between `Sized` and whether the length of a bytestring is known at compile time.

I guess if users were hex-encoding the raw byte patterns of Rust objects (without following any internal pointers) the terms would coincide.

In any case, I'll PR to change the names and we can debate there.

Kixunil:
`Vec` is a `Sized` container for `!Sized` `[u8]`. But I do appreciate that this is a matter of perspective and it might be confusing.

apoelstra:
You can argue that `Box<[u8]>` is a "`Sized` container for `[u8]`", but there's no argument that this is true for `Vec<u8>`. `Vec` is a totally different type with different behavior and different underlying construction.

Kixunil:
They seem extremely similar to me. :shrug:

PR #174, opened 2025-04-23
rename error types of new decoding functions

apoelstra:
The old names were incoherent at best.

### DIFF
--- a/examples/hexy.rs
+++ b/examples/hexy.rs
@@ -9,7 +9,7 @@ use std::fmt;
 use std::str::FromStr;
 
 use hex_conservative::{
-    fmt_hex_exact, Case, DecodeFixedSizedBytesError, DisplayHex as _, FromHex as _,
+    fmt_hex_exact, Case, DecodeFixedLengthBytesError, DisplayHex as _, FromHex as _,
 };
 
 fn main() {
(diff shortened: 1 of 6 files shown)


examples/hexy.rs
```diff
@@ -9,7 +9,7 @@ use std::fmt;
 use std::str::FromStr;
 
 use hex_conservative::{
-    fmt_hex_exact, Case, DecodeFixedSizedBytesError, DisplayHex as _, FromHex as _,
+    fmt_hex_exact, Case, DecodeFixedLengthError, DisplayHex as _, FromHex as _,
```
Kixunil, line 12: It sounds like decoding som kind of length but it's not decoding length, it's decoding bytes. I'd suggest `DecodeFixedLengthBytes` or `DecodeFixedLenBytes`.

Regarding commit message, my main argument is that if the caller then returns the error it "leaks" the type being decoded. E.g. We could later have a function to decode into `Box<[u8]>` so `DecodeToVec` would be misleading.

src/error.rs
```diff
@@ -73,14 +73,14 @@ pub(crate) use write_err;
 /// This error differs from [`DecodeFixedLengthError`] in that the number of bytes is only known
 /// at run time - e.g. when decoding `Vec<u8>`.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum DecodeDynSizedBytesError {
+pub enum DecodeVariableLengthError {
```
Kixunil, line 76: Same issue as in previous commit, sounds like decoding length.

apoelstra:
Added `Bytes` back to the error type names.

apoelstra (commented):
On 658fbf524b26f45fe230e9ba611aa076c4f5f3d6 successfully ran local tests

apoelstra (commented):
On 6e573e809db1724beb75a27b8e18ffeb810f86fa successfully ran local tests

PR #180, opened 2025-06-23
Remove the `HexToBytesIter` from the public API

tcharding:
Face palm! - Thanks goodness for the RC cycle. No clue how this slipped by us.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -29,7 +29,8 @@ impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
     ///
     /// If the input string is of odd length.
     #[inline]
-    pub fn new(s: &'a str) -> Result<Self, OddLengthStringError> {
+    #[allow(dead_code)] // Remove this when making HexToBytesIter public.
+    pub(crate) fn new(s: &'a str) -> Result<Self, OddLengthStringError> {
         if s.len() % 2 != 0 {
             Err(OddLengthStringError { len: s.len() })
         } else {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,6 +65,8 @@ mod iter;
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
 
+use crate::iter::HexToBytesIter;
+
 #[rustfmt::skip]                // Keep public re-exports separate.
 #[doc(inline)]
 pub use self::{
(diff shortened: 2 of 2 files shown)


apoelstra:
@stevenroose we'll put it back once it's stable. The 1.0 release is extremely limited -- it supports only decoding and (importantly) the decoding error types. We were planning to put encoding into 1.1, but if you want `HexToBytesIter` maybe we should do that instead since it's small and probably close to stability, and we can do encoding in 1.2.

The idea here is that users' public APIs will be able to use the 1.0 types (and `FromStr`/`Display`/`LowerHex`/etc) so they can get their own APIs to be stable.

For actual hex functionality they will need to use 0.3.x for a little while. But you don't need to reveal that in your public API so you can quietly replace it with 1.x One Day(TM) when 1.x is ready.

src/iter.rs
```diff
@@ -29,7 +29,8 @@ impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
     ///
     /// If the input string is of odd length.
     #[inline]
-    pub fn new(s: &'a str) -> Result<Self, OddLengthStringError> {
+    #[allow(dead_code)]
```
apoelstra, line 32: In 548aa04953a8c63bb99a83c58a765b7b03966e3e:

We should just delete the function if it's not used.

tcharding: But we will want it back soon as we make this type public, so I figured leave it in there. Right?

tcharding: Added a comment to the attribute.

apoelstra: Eh, I guess this is fine. We'll remember because we talked about it here.

apoelstra (approved):
ACK 7143ac171c74c61ab5502dd241398a9b88355fe1; successfully ran local tests

stevenroose:
> @stevenroose we'll put it back once it's stable. The 1.0 release is extremely limited -- it supports only decoding and (importantly) the decoding error types. We were planning to put encoding into 1.1, but if you want `HexToBytesIter` maybe we should do that instead since it's small and probably close to stability, and we can do encoding in 1.2.
> 
> The idea here is that users' public APIs will be able to use the 1.0 types (and `FromStr`/`Display`/`LowerHex`/etc) so they can get their own APIs to be stable.
> 
> For actual hex functionality they will need to use 0.3.x for a little while. But you don't need to reveal that in your public API so you can quietly replace it with 1.x One Day(TM) when 1.x is ready.

Uff that sounds complicated. I'll see how it turns out, for now we have our 0.3.0 and it works great :)

PR #184, opened 2025-07-03
Remove reexports of low-level error types

Kixunil:
This is done on top of #183 because it might be controversial, so only the last commit is relevant (though we could just merge this one). If this is merged I deem the crate ready for stabilization. However if this approach is rejected I still want to do something about the API duplication before 1.0.

The crate contains a public `error` module but also reexports all of its
content which just makes the API redundant. There were three possible
solutions:

* unpublish the `error` module
* remove all reexports
* remove only some reexports

This commit choses to remove low-level reexports because while it's
plausible that the downstream crates will want to use the high-level
ones in their API, mentioning the low-level ones should be rare.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,33 @@
+# Contributing to hex-conservative
+
+Thanks for the interest in contributing!
+Before you do anything, please familiarize yourself with the stabilization strategy as described in the README, since it places some restrictions on how the code is developed.
+
+## Contributing to the stable crate
+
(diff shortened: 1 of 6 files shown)


apoelstra:
Will let #183 go in first. But concept ACK. I think in general this is the right approach, at least for the top-level of the crate. (For `units::amount` I'm less sure because that's a submodule whose major types are already reexported at the root, so why would a user be looking in the module except to get at the low-level types?).

apoelstra (approved):
ACK 4f188d700f4e11dcaad99a0f3e028abd2b2e8475; successfully ran local tests

PR #186, opened 2025-07-22
Implement top level encode API

tcharding:
Add two functions that do as the name suggests, returning `String`:

- `encode_lower`
- `encode_upper`

Feature gate both on `alloc`.

Implement both by adding a dependency on the pre-1.0 version of `hex-conservative`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,8 @@ pub mod error;
 mod iter;
 
 #[cfg(feature = "alloc")]
+use alloc::string::String;
+#[cfg(feature = "alloc")]
 use alloc::vec::Vec;
 
 use crate::iter::HexToBytesIter;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -109,6 +111,44 @@ pub fn decode_to_array<const N: usize>(hex: &str) -> Result<[u8; N], DecodeFixed
...
+/// each byte in `data` is always encoded using two hex digits. Thus, the resulting string contains
+/// exactly twice as many bytes as the input data.
+///
+/// # Example
+///
+/// ```
+/// use hex_conservative as hex;
+/// assert_eq!(hex::encode_lower("Hello world!"), "48656c6c6f20776f726c6421");
+/// assert_eq!(hex::encode_lower(vec![1, 2, 3, 15, 16]), "0102030f10");
+/// ```
```
tcharding, line 126: I stole all these docs from the [hex](https://docs.rs/hex/latest/hex/) crate - I believe that is within the legality of the licence since it was once said docs are not copyright.

apoelstra: Docs are definitely copyrighted. We cannot pull stuff out of a MIT-licensed crate into a CC0 crate without permission.

apoelstra:
Why have functions that return a `String`? If you just take a `f: &mut fmt::Formatter` then you can implement these without an `alloc` gate.

tcharding:
Aight, i'll have another go. Thanks

PR #188, opened 2025-08-12
Fix deserialization

### LINKED ISSUE #189: Serde deserialization is broken
```rust
#[test]
fn serialize_lower_roundtrip() -> Result<(), serde_json::Error> {
    let bytes: [u8; 4] = [0xdeu8, 0xad, 0xbe, 0xef];
    let serialized: serde_json::Value =
        hex_conservative::serde::serialize_lower(&bytes, serde_json::value::Serializer)?;
    let deserialized: [u8; 4] = hex_conservative::serde::deserialize(serialized)?;
    assert_eq!(bytes, deserialized);
    Ok(())
}
```

This panics during deserialization, due to attempting to deserialize as a map [here](https://github.com/rust-bitcoin/hex-conservative/blob/master/src/serde.rs#L159), instead of deserializing as a `str`.
```rust
    if d.is_human_readable() {
        d.deserialize_map(HexVisitor(PhantomData))
    } else {
        serde::Deserialize::deserialize(d)
    }
```

Fixed in #188

A-Manning:
Fixes #189.

The changes in [src/display.rs](https://github.com/rust-bitcoin/hex-conservative/pull/188/files#diff-aab9bd06633e7680dd7ac54f81eb5007b79dde5dbbde9e0613e040edd766e89a) and [src/iter.rs](https://github.com/rust-bitcoin/hex-conservative/pull/188/files#diff-4bc73dd48e6ccc80311431323609ad12765c3635b53eebe2deaa4be10ecd9be6) were necessary, since using `serde_json` in tests brings an `impl PartialEq<serde_json::Value> for usize` into context, which makes some code in the aforementioned tests ambiguous without explicitly specifying types

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -9,12 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
-name = "dtoa"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5edd69c67b2f8e0911629b7e6b8a34cb3956613cd7c6e6414966dee349c2db4f"
-
-[[package]]
 name = "hex-conservative"
 version = "0.3.0"
 dependencies = [
@@ -51,9 +45,9 @@ checksum = "46dbcb333e86939721589d25a3557e180b52778cb33c7fdfe9e0158ff790d5ec"
 
 [[package]]
 name = "itoa"
-version = "0.3.0"
+version = "1.0.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91fd9dc2c587067de817fec4ad355e3818c3d893a78cab32a0a474c7a15bb8d5"
+checksum = "4a5f13b858c8d314ee3e8f639011f7ccefe71f97f96e50151fb991f267928e2c"
 
 [[package]]
 name = "lazy_static"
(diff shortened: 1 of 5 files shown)


tcharding (approved):
ACK 594983cbd13cfec2274aa27a8bdcc0195ca9ec9e

tcharding:
And thanks for the contribution, and for finding our (my) bugs!

A-Manning:
> To get past CI you'll need to run 'just update-lock-files' and force push please.

Pushed

A-Manning:
> I had play with your tests, perhaps this is cleaner?

IMO it is cleaner for tests to return `Result<(), SomeError>` than to have `.expect("some error message")` throughout. I also prefer to avoid wildcard imports where possible. I'd prefer to keep the tests as-is

apoelstra (approved):
ACK e1fa1d8f80779d2ee9d5ba2e70dbcfe914a41db0; successfully ran local tests

tcharding:
> > I had play with your tests, perhaps this is cleaner?
> 
> IMO it is cleaner for tests to return `Result<(), SomeError>` than to have `.expect("some error message")` throughout. I also prefer to avoid wildcard imports where possible. I'd prefer to keep the tests as-is

Sure, I'm not that bothered but for the record neither of these is done anywhere else in this codebase (or the `rust-bitocin` org) as far as I'm aware.

apoelstra:
Huh. I actually didn't notice the `Result` return type here, nor did I know that it was allowed (since Rust 2018, apparently). I kinda prefer it.

apoelstra:
As for avoiding wildcards, in the specific case of `super::*` in inline modules, I don't mind them, because any symbols it exports you know are defined in the same file that you're already looking at.

But agreed in general it's better to avoid them.

I'm gonna go ahead and merge this, and if we want to do some broader unit test refactoring effort, we can do that separately.

tcharding:
Backported to `0.3.x` in #201 and to `0.2.x` in #202

PR #191, opened 2025-10-07
Fix lints and remove `doc_auto_cfg`

tcharding:
`rust-bitcoin` is failing to lint and docs build with a recent nightly toolchain because of this crate.

Fix both issues then update nightly to prove its ok.


apoelstra (approved):
ACK b79a0e1db94adb2855f6dfbde29d6b3f80750e62; successfully ran local tests

tcharding:
Ah true that, I'll do a backport then we can release either or both?

tcharding:
Shall we give this a few days and see what happens in the community first?

PR #199, opened 2026-01-28
Release tracking PR: `hex-conservative 0.3.2`

nyonson:
Cherry picks over the const hex! macro so that rust-bitcoin can drop the hex-lit dependency.

Obviously low priority, but would be easy cleanup if released.


apoelstra (approved):
ACK 54ec5a43f62da406e7791ccffcff0081584da1f4; successfully ran local tests

PR #203, opened 2026-02-24
Make `iter` module public

tcharding:
Do a couple of preparatory patches then make the module pubic as well as the `HexToBytesIter::new` constructor. With this applied exactly two types will be added to the public API: `HexToBytesIter` and the helper `HexDigitsIter`. The former can only be constructed using `new` and the latter does not provide a constructor. It is meant for usage only if one needs to refer to the iter type explicitly as `HexToBytesIter<HexDigitsIter>`.

Note we do not add the `HexSliceToBytesIter` alias from master. If required that can be done as a follow up because as is this PR only touches code on the 1.x branch already.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -29,8 +29,7 @@ impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
     ///
     /// If the input string is of odd length.
     #[inline]
-    #[allow(dead_code)] // Remove this when making HexToBytesIter public.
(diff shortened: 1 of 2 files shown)


tcharding:
Woops wrong base.

apoelstra (approved):
ACK 508962bd30b70c5cce65971b1e69d4de2c4bba06; successfully ran local tests; yep, this API seems totally reasonable to me.

PR #206, opened 2026-02-27
Prepare master to be `v1.1` 

tcharding:
The alarming diffstat is caused by the API text files.

Full disclosure; I had PTSD from 1.0-ing this crate, I should have prepared the `v1.1` release months and months ago - my bad. Turns out after I made the creative leap to copy the code back from 1.0 instead of the other way around this wasn't that hard.

The bit I'm patting my  self on the back about is the last two patches: (1) create API text files on the `1.x` branch and commit them on master (as well as a script). (2) run the script again, resulting in only green lines - WIN.

The controversial parts of this PR are:

- Delete the `serde` module - boo hoo.
- Delete the `FromHex` trait. 

Before release we will want to go over the whole crate with a fine tooth comb. ~I propose doing `v0.4.0` as an RC release for `v1.1`~.

EDIT: Consensus in the discussion below is to go straight for `v1.1.0`.


tcharding:
TL;DR lets just release `v1.1` with no release candidate and be done with this whole fucking `hex` business.

> > We can't semver 1.0 types on master and then release master as 1.1
> 
> We never need to "semver trick" 1.0 types into 1.1. Cargo will always unify the minor versions of 1.x across the dependency tree.

Yep yep, but `master` is going to become `0.4.0` and then, hopefully with no changes, `v1.1`. I don't want to semver on master (delete all the `v1.0.0.` code), then re-introduce all the code again to do the `v1.1.0` release. See what I mean?
 
> > Am I on drugs; you got the `git diff` arguments around the wrong way, right?
> 
> Ah, yes, you're right. Okay, I see what you're trying to do here.
> 
> > * Get rid of the unstable and stable hex deps, replace with `hex v0.4` (this is a breaking change, will require new RC for whole stack)
> 
> ...yes, and it will also mean that nobody can use our crates alongside `hex 1.0` because the error types will not be compatible.

What about we merge this. Gather the whole team together and turn the crank on master (API checklist, docs, etc)  for a week (or two ...). Then release the crate with no RC release, dust our hands, and just say, "that's the best we can do and fuck you if you don't like it"? The RC is not going to catch anything because there is basically no change to `0.3` and we have been using that for eons. 

> I think if we released `1.1-rc` that would do the right thing, but I would need to check. (The "right thing" is that cargo updates every instance of hex `1.0` in the dep tree to `1.1-rc`.)

I personally have no faith in `-rc` 'doing the right thing'.

apoelstra (approved):
ACK 5da30306267031e37226d8262c61de36bbee634d; successfully ran local tests

apoelstra (approved):
ACK 56a74fed23a09203b05beb88f785f7112be2d582; successfully ran local tests

PR #211, opened 2026-04-02
Re-write CI to use `cargo rbmt`

tcharding:
Copy what we do in `rust-bitcoin` and re-write CI to use `cargo-rbmt`. Note a couple of the changes I don't remember why we did, git log explicitly states these. 

I hazard a guess that CI wouldn't be green on each individual patch; patch separation for the benefit of review.

I can't work out why `contrib/check-for-api-changes.sh` and `cargo rbmt api` produce different output. Locally I have `cargo-public-api 0.50.1` installed. And I don't know if this means the 'prove we did the copy from 1.0.x branch correctly' is still valid with this API text file confusion?


.github/workflows/rust.yaml
```diff
@@ -9,174 +8,62 @@ on:   # yamllint disable-line rule:truthy
...
-    runs-on: ubuntu-latest
+  Test:
+    name: Test - ${{ matrix.toolchain }} toolchain, ${{ matrix.dep }} deps
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
     strategy:
       fail-fast: false
       matrix:
+        toolchain: [stable, nightly, msrv]
```
nyonson, line 19: Just noting testing `nightly` here probably not worth it, talked about it over in psbt: https://github.com/rust-bitcoin/rust-psbt/pull/87#issuecomment-4151565331

nyonson:
There are a least 2 things causing API file changes.

1. The significant nightly toolchain version bump causes the underlying docs to shift around.
2. The legacy `check-for-api-changes.sh` script in this repo uses public-api's `--simplified` flag which omits blanket impls. cargo-rbmt's usage could be updated to do the same if we prefer this ([there are also some other omit settings](https://docs.rs/public-api/latest/public_api/struct.Builder.html#method.omit_blanket_impls)):

```rust
        let public_api = public_api::Builder::from_rustdoc_json(&json_path)
            .omit_blanket_impls(true)
            .build()?;
```

I am not sure yet if that is an exhaustive list of reasons for the API diff though.

tcharding:
I can work with that info, thanks man!

apoelstra:
Please highlight me once you're comfortable with the state of the API files, and this is ready for review.

tcharding:
I convinced myself that this PR is correct by doing #217 (demo not merge candidate). 

This is right to go @apoelstra

tcharding:
Force push is just re-write of the git log of commit: `5a4a7bf Update API text files`

apoelstra:
In e32f150f0a8e324416769b8367f03c02d14392b5:

FYI the "why change the Github runner" it's because sometimes Github updates the runners and breaks our stuff.

Though conversely, if we pin the runner, then eventually Github will delete the runner we're using which will also break our stuff. So there is really no good solution here.

apoelstra:
Hah, nice, so this is the PR where you force me to actually update to use the new `nightly-version` location..

apoelstra (approved):
ACK 7b47081f725035c9fa1d9522a417a46ebe789837; successfully ran local tests

apoelstra:
Gonna prioritize this one over #213 #219 #220, sorry, since those are easier to re-review.

PR #215, opened 2026-04-10
Convert DisplayHex to use GAT

mpbagot:
Since it's inception, the DisplayHex trait should ideally have made use of a GAT for the Display associated type. Due to limitations in early Rust versions, this was not possible. By bumping the MSRV to 1.74.0 in the new version of the crate, the DisplayHex trait can now be redesigned using a GAT for the associated Display type. This allows for the function signature of the as_hex to be changed to suit the expected naming convention.

- Patch 1 changes the DisplayHex trait to use a GAT for the Display type, adjusting implementations.
- Patch 2 removes the redundant DisplayHex impls for arrays and Vec. Also removes the DisplayArray type.
- Patch 3 updates the API files.

### DIFF
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -49,13 +49,12 @@ where
 {
     a: BytesToHexIter<I>,
     b: buf_encoder::BufEncoder<CAP>,
-    c: display::DisplayArray<'a, CAP>,
-    d: display::DisplayByteSlice<'a>,
+    c: display::DisplayByteSlice<'a>,
     #[cfg(feature = "std")]
-    e: display::HexWriter<T>,
-    f: Case,
-    g: HexToBytesIter<J>,
-    h: HexSliceToBytesIter<'a>,
+    d: display::HexWriter<T>,
+    e: Case,
+    f: HexToBytesIter<J>,
+    g: HexSliceToBytesIter<'a>,
     _marker: PhantomData<T>, // For when `std` is not enabled.
 }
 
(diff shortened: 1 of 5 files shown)


apoelstra:
**Edit** the below text applies to an earlier version of this PR which had a rename and no MSRV bump.

Lol, this is going to piss off everybody who replaced `to_hex` with `as_hex` 100 times the last time we made this break.

**However** in this case I think we're fixing the wrong problem. I believe that this method takes `self` because without GATs, it's impossible to take `&self` and return an associated type. And then it has a `Copy` bound to discourage people from implementing this trait in a way that'd force their users to call `Clone` a bunch.

This all dates to https://github.com/rust-bitcoin/rust-bitcoin/pull/1268 which introduced these types -- and in that thread Kix says exactly this ("if I had GATs I'd...").

I think we should just bump the MSRV and use GATs here.

apoelstra:
Nice. So, Clippy is complaining about a couple other things this MSRV has unlocked.

Do we actually need a `DisplayHex` impl for `Vec` and all the array types now? They should deref to `[u8]` and then use the impl on `[u8]` now, right?

Also, I would suggest unsealing the trait now.

tcharding:
I did #216 because I felt like doing something light and thought it might be a help to you @mpbagot. But I got no local lint warnings ...

mpbagot:
I rebased on #216, removed the seal and the Vec/array impls. That also removes DisplayArray entirely.
I don't see any lint errors locally, so I still haven't changed anything else.

apoelstra:
6f9a2c5ba1552e9a24c73c4b87bd563c0566255b needs rebase

tcharding (approved):
ACK 6f9a2c5ba1552e9a24c73c4b87bd563c0566255b

mpbagot:
rbmt is installing 1.63 for the MSRV toolchain still.
Edit: The Cargo.toml MSRV wasn't actually bumped in #216, just everything around it.

tests/api.rs
```diff
@@ -98,15 +96,13 @@ fn api_all_non_error_types_have_non_empty_debug() {
     assert!(!debug.is_empty());
     let debug = format!("{:?}", t.c);
     assert!(!debug.is_empty());
+    #[cfg(feature = "std")]
     let debug = format!("{:?}", t.d);
     assert!(!debug.is_empty());
-    #[cfg(feature = "std")]
     let debug = format!("{:?}", t.e);
     assert!(!debug.is_empty());
```
tcharding, line 110: The feature gating here is buggy before and after this change.

Highlighted in: #231

tcharding: Actually its not buggy, just misleading, my bad. Still better to use paranthesis I rekon

```rust
  #[cfg(feature = "std")]
  {
      let debug = format!("{:?}", t.e);
      assert!(!debug.is_empty());
  }
```

tcharding: This is going to conflict with #231 anyways so probably you can just ignore me for now.

tcharding (approved):
ACK 34849ab02f36ac2825000f277da5980debec3076

apoelstra (approved):
ACK 34849ab02f36ac2825000f277da5980debec3076; successfully ran local tests

PR #216, opened 2026-04-12
Bump MSRV to 1.74.0

tcharding:
Bump the MSRV. Then update the nightly toolchain because there are lint warnings because we have the new clippy config option `empty_enums` already.


README.md
```diff
@@ -28,14 +28,15 @@ by `std`.
...
 reserve the right to use features to guard compiler specific code so `--all-features` may not work
 using the MSRV toolchain.
 
 ### Policy
 
-We don't intend to bump MSRV until the newer Rust version is at least two years old and also
-included in Debian stable (1.63 is in Debian 12 at the moment).
+We don't tend to bump MSRV until the newer Rust version is at least two years old and also included
+in Debian stable. Debian 13 (trixie) ships with 1.85 however for `rust-bitcoin` and down the stack
+we elected to stick with Rust 1.74.
```
apoelstra: In e369108c77c5bf60aa773b2d2eeb1fb7019ecc3d:

I don't like this new text. I don't know what "We don't tend to bump MSRV" means, and the extra comments about rust-bitcoin feel like internal details. I think we should just say that we *may* bump our MSRV in a minor version, but we have no plans to, and that we promise the next MSRV will be at least 2 years old and shipped in Debian stable.

tcharding: Will we? Why would we bump it in a minor version and not do a 2.0 and semver (so we do it properly)?

tcharding: I used: "Policy is to never use an MSRV that is less than two years old and also that ships in Debian stable."

apoelstra:
afca9f427416f74217416877c752465b700e1a68 needs rebase

README.md
apoelstra: @tcharding we're doing a MSRV bump in a minor release now. Presumably the same conditions might eventually hold.

The semver trick might also not work. Suppose we want to do a MSRV bump to implement `core::error::Error` on our error types. This change would break almost every type and function in the library, in a non-semver-trickable way.

> Will we?

Realistically probably not. But we *may*.

tcharding: No worries, added "We may bump our MSRV in a minor version, but we have no plans to. "

apoelstra (approved):
ACK 97e0e82019c4199ec83fffca986cac1672268af5; successfully ran local tests

apoelstra (approved):
ACK 3ff2412ec5736c08b573bcebde7f4e2f5577b654; successfully ran local tests

tcharding:
Face palm - done correctly in #228.

tcharding:
My mistake was that my grep alias excludes `Cargo.toml` - fixed.

tcharding:
Nope, turns out I'm just a goose ...

PR #217, opened 2026-04-12
DEMO: Check that we did the migration correctly

tcharding:
#211 highlighted that `cargo rbmt api` produced different output from the custom API script we used during the migration.

In order to prove that we didn't make mistakes I did this PR which does not need to be merged. 

Steps described in the commit log of patch 1.


tcharding:
Can be closed now #211 merged.

PR #218, opened 2026-04-12
Do not inline error types into docs

tcharding:
As we have recently started doing in `rust-bitcoin`; do not inline error types because that way the docs render nicely with the errors in a different section to the rest of the public types.


apoelstra (approved):
ACK f58a83bb2aa02abe673bcb2f29760faaf94459c4; successfully ran local tests

PR #219, opened 2026-04-12
Do trivial error cleanups

tcharding:
As we just did in `rust-bitcoin` move the `From<Infallible>` impl and be explicit when implementing `error::Error`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -246,7 +246,10 @@ impl fmt::Display for InvalidCharError {
 }
 
 if_std_error! {{
-    impl StdError for InvalidCharError {}
+    impl StdError for InvalidCharError {
+        #[inline]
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 5d5e3b8b9ae3e5582ed0ccf14db9febcb92b29f1; successfully ran local tests

tcharding:
Oh this introduced API text file changes because default impl of `std::error::Error` does not show up in the files. My bad, clearly we do not have a CI job that checks it but I'm inclined to leave that as is for this week as we push hard on `v1.1.0`, @apoelstra?

Fixed in #222

apoelstra:
We should be okay from here on out (after merging 222), right, because we've moved to cargo-rbmt?

tcharding:
With this we will: https://github.com/rust-bitcoin/hex-conservative/pull/226

PR #220, opened 2026-04-12
Derive `Clone`

tcharding:
Seems to be no reason not to derive `Clone` on all types that we can. `grep` for `derive(Debug)` and update all sites to also derive `Clone`.

PR #221, opened 2026-04-12
Derive standard traits on all types

mpbagot:
According to the C-COMMON-TRAITS API guideline, types in Rust libraries should implement common std lib traits where possible, due to orphan rule restrictions preventing downstream users from doing so easily.

Add derives for Debug, Clone, PartialEq, Eq and Hash traits to all types.

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -221,6 +221,7 @@ impl DisplayHex for [u8] {
 /// Displays byte slice as hex.
 ///
 /// Created by [`<&[u8] as DisplayHex>::as_hex`](DisplayHex::as_hex).
+#[derive(Clone, PartialEq, Eq, Hash)]
 pub struct DisplayByteSlice<'a> {
     // pub because we want to keep lengths in sync
     pub(crate) bytes: &'a [u8],
@@ -493,7 +494,7 @@ where
 /// Given a `T:` [`fmt::Write`], `HexWriter` implements [`std::io::Write`]
 /// and writes the source bytes to its inner `T` as hex characters.
 #[cfg(feature = "std")]
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct HexWriter<T> {
     writer: T,
     table: &'static Table,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,7 +225,8 @@ mod table {
     /// Table of hex chars.
     //
     // Correctness invariant: each byte in the table must be ASCII.
-    #[derive(Debug)]
+    #[allow(clippy::derived_hash_with_manual_eq)] // The Eq impl distinguishes the two possible values of Table
+    #[derive(Debug, Hash)]
     pub(crate) struct Table([u8; 16]);
 
     impl Table {
@@ -261,6 +262,13 @@ mod table {
             hex_str
         }
     }
+
+    impl PartialEq for Table {
+        // Table can only be Table::LOWER or Table::UPPER. These differ in any of the bytes from
+        // indices 10-15.
+        fn eq(&self, other: &Self) -> bool { self.0[10] == other.0[10] }
+    }
+    impl Eq for Table {}
 }
 
 #[cfg(test)]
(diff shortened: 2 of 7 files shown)


tcharding (approved):
ACK 4aba70c36bcc64a3f134cd974cd9b34a4dfee023

tcharding:
Lets do this instead of my #220

src/display.rs
```diff
@@ -609,7 +611,7 @@ where
 /// Given a `T:` [`fmt::Write`], `HexWriter` implements [`std::io::Write`]
 /// and writes the source bytes to its inner `T` as hex characters.
 #[cfg(feature = "std")]
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
Kixunil, line 497: These traits look super weird to me. It's as if they were implemented on a pipe or something. I think `Clone` makes some sense but I'm not comfortable with defining equality. The type should be short-lived anyway.

src/lib.rs
```diff
@@ -225,7 +225,7 @@ mod table {
     /// Table of hex chars.
     //
     // Correctness invariant: each byte in the table must be ASCII.
-    #[derive(Debug)]
+    #[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
Kixunil: This is not public API and thus these traits are just waste of compile time.

Kixunil (dismissed):
Please remove the useless traits and think for a minute about the questionable ones. (I'm not hard against those but better double check.)

src/display.rs
apoelstra: If the underlying writer is clonable or comparable, why should wrapping it in hex change that?

Kixunil: We have to commit to the comparison behavior. What makes writers the same and what makes them different? Is different `Table` a different writer?

`Clone` isn't that weird from behavior standpoint but if you're cloning this kind of writer you're doing something very weird. Anyway, I don't really object to it, just pointing it out.

src/lib.rs
mpbagot: They're needed if we want the derives on `HexWriter`, `BufEncoder`, or `BytesToHexIter`, since all of those hold an `&Table`. If we drop the derive traits on those types, then this can go away.
`Clone` can be removed regardless.

src/display.rs
apoelstra: > We have to commit to the comparison behavior. What makes writers the same and what makes them different? 

Well, we pass that question up to the underlying writer.

> Is different `Table` a different writer?

It's a different `HexWriter`, yes. I agree that we're committing to this behavior and it's hard to say what it "really means", but it's also intuitively what you would expect: if the two underlying writers are the same, then we compare tables, and lowercase is lower than uppercase.

Kixunil: OK, I don't mind this then.

src/lib.rs
Kixunil: Ah, I see. But in such case I would prefer implementing those traits in terms of comparing identity rather than trying to compare the entire table. We only ever have two tables: lower and upper, so it's best to compare pointers instead. And yes, please remove `Clone`.

apoelstra:
3168555a1305380084699dda0d059620678ecf9a needs rebase

apoelstra (approved):
ACK 816da38a971f0daa4520c5591eb21b984bde425c; successfully ran local tests

tcharding (approved):
ACK 816da38a971f0daa4520c5591eb21b984bde425c

apoelstra:
Feel free to file a followup issue about `Hash`.

src/lib.rs
```diff
@@ -261,6 +262,13 @@ mod table {
             hex_str
         }
     }
+
+    impl PartialEq for Table {
+        // Table can only be Table::LOWER or Table::UPPER. These differ in any of the bytes from
+        // indices 10-15.
+        fn eq(&self, other: &Self) -> bool { self.0[10] == other.0[10] }
```
Kixunil, line 269: Nice trick and arguably less error-prone than comparing the pointer, so I'm fine with it.

I'll be happy to accept a similar `Hash` impl if someone feels like it.

PR #224, opened 2026-04-13
Remove serde dependencies

mpbagot:
With the removal of the serde module from the 1.1 release, the serde and serde_json dependencies are no longer needed at all. In case users build with all features, we should remove them from the dependency list to prevent the needless compilation of unused dependencies.

Remove serde and serde_json from dependencies and dev-dependencies.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -14,8 +14,6 @@ version = "0.3.0"
 dependencies = [
  "arrayvec",
  "if_rust_version",
- "serde",
- "serde_json",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


tcharding (approved):
ACK 6bbfd58089a4a72fad519d687a837775ece65dcc

tcharding:
Lucky catch! This is not breaking because `serde` feature did not exist in the `v1.0.0` release.

apoelstra:
Agreed, great catch!

apoelstra (approved):
ACK 6bbfd58089a4a72fad519d687a837775ece65dcc; successfully ran local tests

PR #225, opened 2026-04-13
Add Case and HexToBytesIter to api test

mpbagot:
As per the C-DEBUG API guideline, all public types must have a Debug impl. While all such types do, the presence and non-emptyness of the impls on Case and HexToBytesIter should be tested in the api test to prevent regressions.

Add Case and HexToBytesIter (via HexSliceToBytesIter) to the api debug test.


tcharding (approved):
ACK 99469c06478b6e98a36f373f93fd8f8d33e97326

apoelstra (approved):
ACK 99469c06478b6e98a36f373f93fd8f8d33e97326; successfully ran local tests

PR #223, opened 2026-04-12
Add `FromIterator` and `Extend` to `BufEncoder`

mpbagot:
The BufEncoder type represents a collection of bytes. According to the C-COLLECT API guideline, collections should implement FromIterator and Extend in order to allow for simple construction using functions like collect() on iterators.

Implement FromIterator and Extend on BufEncoder.

PR #222, opened 2026-04-12
Update API files

mpbagot:
The current state of the crate is not reflected in the public API files on master. They should be updated.

PR #226, opened 2026-04-13
CI: Add api job

tcharding:
I thought it might be better not to do this in order to not slow us down and cause rebases and I'm still not 100% what is better.

Elect to take the safe but possibly annoying route and enable the job.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -33,7 +33,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


apoelstra:
#222 is merged. You should be able to rebase this and it'll pass.

apoelstra (approved):
ACK d0af85049bfa4c8ac33f37ff2a69eda1a4f8094d; successfully ran local tests

apoelstra:
Merged concurrently with #223 #224 and #225, so apologies if I broke the new job because those other PRs had incorrect API files.

PR #227, opened 2026-04-13
Convert `HexSliceToBytesIter` to a newtype instead of an alias

mpbagot:
The HexSliceToBytesIter type is currently a type alias. This reveals representation and implementation details of the type to the public API which limits the potential for changes in the future. Instead, this type should be implemented as a newtype wrapper around the concretised generic type.

Replace HexSlicetoBytesIter alias with a newtype tuple struct.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -14,8 +14,58 @@ use crate::alloc::vec::Vec;
 use crate::error::{InvalidCharError, OddLengthStringError};
 use crate::{Case, Table};
 
-/// Convenience alias for `HexToBytesIter<HexDigitsIter<'a>>`.
-pub type HexSliceToBytesIter<'a> = HexToBytesIter<HexDigitsIter<'a>>;
+/// Iterator over bytes decoded from a hex string slice.
+///
+/// This is an iterator type returned when decoding a `&str` of hex digits. Each pair of hex
+/// characters is decoded into one byte.
+///
+/// Use [`HexToBytesIter`] if you need an iterator that is generic over the source of hex digit
+/// pairs.
+#[derive(Debug)]
+pub struct HexSliceToBytesIter<'a>(HexToBytesIter<HexDigitsIter<'a>>);
+
+impl<'a> HexSliceToBytesIter<'a> {
+    /// Constructs a new [`HexSliceToBytesIter`] from a string slice.
+    ///
+    /// # Errors
+    ///
+    /// If the input string is of odd length.
+    #[inline]
+    pub fn new(s: &'a str) -> Result<Self, OddLengthStringError> {
+        HexToBytesIter::new(s).map(Self)
+    }
+}
+
+impl Iterator for HexSliceToBytesIter<'_> {
+    type Item = Result<u8, InvalidCharError>;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> { self.0.next() }
+
+    #[inline]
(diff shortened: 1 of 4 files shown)


src/iter.rs
```diff
@@ -34,7 +84,7 @@ impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
     ///
     /// If the input string is of odd length.
     #[inline]
-    pub fn new(s: &'a str) -> Result<Self, OddLengthStringError> {
+    pub(crate) fn new(s: &'a str) -> Result<Self, OddLengthStringError> {
```
tcharding, line 87: Why this change mate?

tcharding:
Everything else looks good: 816e4b042e6452ec7f7c47e74c2f01ed1a1b8d49

src/iter.rs
mpbagot: The idea was to push users into using the newtype instead of manually typing out the concrete type. If this is left pub, they can construct either with the same constructor signature. Not a critical change, I was just thinking that it's always easier to add it back in later than remove it if it turns out we want to force the newtype usage.

tcharding: Without this no one can construct a generic one though, right?

tcharding:
Can you patch the `api.rs` file as well with this mate?

src/iter.rs
```diff
@@ -14,8 +14,58 @@ use crate::alloc::vec::Vec;
 use crate::error::{InvalidCharError, OddLengthStringError};
 use crate::{Case, Table};
 
-/// Convenience alias for `HexToBytesIter<HexDigitsIter<'a>>`.
-pub type HexSliceToBytesIter<'a> = HexToBytesIter<HexDigitsIter<'a>>;
```
tcharding, line 18: I wonder if we should make the `HexDigitsIter` private now? I believe it was only public so it could appear in this type alias.

src/iter.rs
mpbagot: There are two constructors on `HexToBytesIter`:
- `new` constructor is specifically for `HexToBytesIter<HexDigitsIter<'a>>`
- `from_pairs` constructor is for `HexToBytesIter<I: Iterator<Item = [u8; 2]>>`

This change will mean that the `new` constructor will now only be accessible on this newtype, while the `from_pairs` will remain on the generic type.

src/iter.rs
mpbagot: In the rendered docs, it's not even clickable. It's pub only in the private module `iter`, so I don't think that it's actually possible to directly access at the moment anyway. This PR will completely indirectly remove it from the public API, unless we later make the `iter` module pub.

mpbagot:
> Can you patch the api.rs file as well with this mate?

#225 already put it in the `api.rs` test file, this is just slightly behind master since that merged.

src/iter.rs
tcharding: Oh I missed `from_pairs` in the different impl block. My bad. Note that `from_pairs` requires `ExactSizedIterator`. But anyways, your change is correct because one cannot use `new` anyway because `HexDigitsIter` is not public available. So this is an API fix really.

tcharding (approved):
ACK 816e4b042e6452ec7f7c47e74c2f01ed1a1b8d49

apoelstra (approved):
ACK 816e4b042e6452ec7f7c47e74c2f01ed1a1b8d49; successfully ran local tests

PR #229, opened 2026-04-13
Document the hex macro

tcharding:
This is mainly just a test macro but we should still document it thoroughly.


apoelstra (approved):
ACK 41c4a9c4c92dca4e033bd9547d839a62c262b983; successfully ran local tests

PR #230, opened 2026-04-13
Implement perf optimization using `MaybeUninit`

### LINKED ISSUE #146: Use `MaybeUninit` in place of 0-initialized u8 arrays
_Originally posted by @Kixunil (in a particular case) in https://github.com/rust-bitcoin/hex-conservative/pull/123#discussion_r1936918425_

There may be other cases -- typically with code Tobin or I touched we were not looking for unsafe optimizations.

tcharding:
Using a zero initialized array on the stack may be slower than using a `MaybeUninit` because of the required initialization to zero just to overwrite the buffer.

However using (testing) `MaybeUninit` is a bit of a nuisance so leave the original function `drain_to_slice` in the codebase but cfg it out unless we are testing.

Note to review: This is a bit smelly, feel free to reprimand me.

Close #146

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -5,6 +5,7 @@
 use core::borrow::Borrow;
 use core::convert::TryInto;
 use core::iter::FusedIterator;
+use core::mem::MaybeUninit;
 use core::str;
 #[cfg(feature = "std")]
 use std::io;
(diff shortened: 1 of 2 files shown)


tcharding:
I wrote the code with an Opus 4.6 (all the smelly bits, duplicate test function etc was all me).

src/lib.rs
```diff
@@ -137,10 +138,20 @@ pub fn decode_to_vec(hex: &str) -> Result<Vec<u8>, DecodeVariableLengthBytesErro
 /// `N * 2`.)
 pub fn decode_to_array<const N: usize>(hex: &str) -> Result<[u8; N], DecodeFixedLengthBytesError> {
     if hex.len() == N * 2 {
-        let mut ret = [0u8; N];
+        // SAFETY: `[MaybeUninit<u8>; N]` has no initialization requirement,
+        // so an uninitialized array of them is sound. This is the standard
+        // `uninit_array` pattern.
+        let mut ret: [MaybeUninit<u8>; N] = unsafe { MaybeUninit::uninit().assume_init() };
```
apoelstra, line 144: In 17054126e8bf803f52ba44a89d54ebcf766e9344:

Can you find a citation for this being the "standard `uninit_array` pattern"? It appears intuitively that `MaybeUninit::uninit().assume_init()` is always unsound.

AFAICT, by reading for example [this rust-lang thread](https://github.com/rust-lang/rust/pull/101179), the correct way to do this on stable Rust is `[const { MaybeUninit::uninit() }; N]`.

src/iter.rs
```diff
@@ -47,14 +48,39 @@ impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
...
+        let mut ptr = buf.as_mut_ptr().cast::<u8>();
+        for byte in self {
+            // SAFETY: for loop iterates `len` times, and `buf` has length `len`.
+            // Writing through a `*mut u8` derived from `*mut MaybeUninit<u8>` is sound
+            // because the two have identical layout and `MaybeUninit<u8>` imposes no
+            // validity requirement on the bytes being overwritten.
+            unsafe {
+                core::ptr::write(ptr, byte?);
+                ptr = ptr.add(1);
+            }
```
apoelstra, line 76: In 17054126e8bf803f52ba44a89d54ebcf766e9344:

Why use `ptr::write` instead of `Maybeuninit::write`? That would save you unsafety, casts, and pointer arithmetic.

apoelstra:
>I wrote the code with an Opus 4.6 

Thanks for the disclosure. From a new contributor this would be a "close without bothering to comment" PR.

tcharding:
Ouch, I guess I should not to vibe code issues that I don't care about. Sorry for using your time bro.

tcharding:
Drafting because for me personally this is not high priority.

src/lib.rs
Kixunil: > It appears intuitively that `MaybeUninit::uninit().assume_init()` is always unsound.

I don't have a citation but that thing is not unsound when the resulting type consists of only `MaybeUninit`. I've seen that pattern being used in a way that suggests it's correct.

src/iter.rs
Kixunil: or even `MaybeUninit:new` if `write` is below MSRV.

src/iter.rs
```diff
@@ -47,14 +48,39 @@ impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
...
     ///
     /// # Panics
     ///
     /// Panics if the length of this `HexToBytesIter` is not equal to the length of the provided
     /// slice.
+    pub(crate) fn drain_to_uninit_slice(
+        self,
+        buf: &mut [MaybeUninit<u8>],
+    ) -> Result<(), InvalidCharError> {
+        // IF YOU CHANGE THIS FUNCTION DO THE ONE BELOW TOO.
```
Kixunil: I believe the one below will be completely replaceable by this one. While technically this one should be `drain_to_out`, because it also promises initialization, I think it's OK to promise no de-initialization.

PR #231, opened 2026-04-13
Cover HexToBytesIter in api test

jamillambert:
Include `HexToBytesIter` in the API surface test and fix a feature gating of an assert.

### DIFF
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -101,8 +101,10 @@ fn api_all_non_error_types_have_non_empty_debug() {
     let debug = format!("{:?}", t.d);
     assert!(!debug.is_empty());
     #[cfg(feature = "std")]
-    let debug = format!("{:?}", t.e);
-    assert!(!debug.is_empty());
+    {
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK ff58058a19347ebd7bf6573b9ea48b844ecf376b

apoelstra (approved):
ACK ff58058a19347ebd7bf6573b9ea48b844ecf376b; successfully ran local tests; though I'm not sure how we didn't catch the `std` thing in either Github or local CI?

tcharding:
The code is syntactically correct before the patch just not semantically correct (we re-use `debug` local var so the assertion is checking the previous setting of it).

EDIT: Oh did you mean Andrew that you were surprised none of our tools picked up on asserting twice on the same var (without setting it when `std` wasn't enabled)?

apoelstra:
Ah, no I meant the first thing. I didn't realize that we were just picking up an old value of the variable.

PR #232, opened 2026-04-13
Implement `nth` and `nth_back`

### LINKED ISSUE #14: Implement `nth` and `nth_back`
Override the default implementations for hex iterators.

jamillambert:
Overriding `nth` and `nth_back` can be O(1) instead of the current O(n).

- Override the default implementations.
- Add tests for the new implementations.
- Add a benches crate with benchmarks comparing the new implementations to the old ones.
- Update API files

Benchmark results:
| Operation | Size (bytes) | Optimized (median) | Linear baseline (median) | Speed increase |
|---|---:|---:|---:|---:|
| BytesToHexIter::nth | 128 | 6.73 ns | 191.05 ns | 28.4× |
| BytesToHexIter::nth | 4096 | 6.57 ns | 5.615 µs | 854.5× |
| BytesToHexIter::nth_back | 128 | 6.94 ns | 163.65 ns | 23.6× |
| BytesToHexIter::nth_back | 4096 | 7.12 ns | 4.754 µs | 667.8× | 

Closes #14

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ exclude = ["api", "tests", "contrib"]
 
 [workspace]
 members = ["fuzz"]
+exclude = ["benches"]
 
 [features]
 default = ["std"]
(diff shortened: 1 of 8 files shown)


apoelstra:
Holy shit @ these benchmark results.

I wonder if this would be meaningfully simpler (or faster) if we passed through to `nth` or `nth_back` on the underlying iterator. Perhaps not, because we have to deal with the cached low nybble no matter what. So ACK this. I reviewed the logic carefully.

apoelstra (approved):
ACK c965c1c120d43217fc583d1b5180d4e604b01182; successfully ran local tests

PR #233, opened 2026-04-14
Do trivial improvements to API surface test

tcharding:
Just to double check the test I went over it super closely. I'm happy with it _not_ having all the re-export tests that we do over in `units` and `primitives` because I'm starting to think they don't add much value now anyways.

All I found were two trivial improvements. Done to ensure we can check off item in #209 after this and #231 go in.


apoelstra (approved):
ACK 32c23b8bf803cc75a4e799ae0b90e3b8c71e907a; successfully ran local tests

PR #235, opened 2026-04-15
Fix `InvalidCharError` position with mixed reading on `HexToBytesIter`

mpbagot:
When mixing next() and next_back() calls on a single HexToBytesIter, the logic used to track the current character position calculates incorrect values. When returning an InvalidCharError, this results in the wrong position being provided to users, which is confusing.

Adjust HexToBytesIter to hold forwards read position to fix current character read position in error case.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -74,7 +74,8 @@ where
     I: Iterator<Item = [u8; 2]>,
 {
     iter: I,
-    original_len: usize,
+    /// Number of (high, low) char pairs consumed from the front.
+    front_pos: usize,
 }
 
 impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
@@ -149,7 +150,7 @@ where
 {
     /// Constructs a custom hex decoding iterator from another iterator.
     #[inline]
-    pub fn from_pairs(iter: I) -> Self { Self { original_len: iter.len(), iter } }
+    pub fn from_pairs(iter: I) -> Self { Self { front_pos: 0, iter } }
 }
 
 impl<I> Iterator for HexToBytesIter<I>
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 6ddebcee2a162c1b11a5b25e125acefd5b8bfa75

src/iter.rs
```diff
@@ -159,31 +160,19 @@ where
...
+    fn next(&mut self) -> Option<Self::Item> { self.nth(0) }
 
     #[inline]
     fn size_hint(&self) -> (usize, Option<usize>) { self.iter.size_hint() }
 
     #[inline]
     fn nth(&mut self, n: usize) -> Option<Self::Item> {
         let [hi, lo] = self.iter.nth(n)?;
+        let pos = (self.front_pos + n) * 2;
+        self.front_pos += n + 1;
```
apoelstra: In 6ddebcee2a162c1b11a5b25e125acefd5b8bfa75:

Both of these lines can be made to panic by providing large input values (or low ones, but calling repeatedly).

tcharding: Ouch, failing reviews in public hurts.

mpbagot: Because L170 will early exit with a `None` if n goes out of bounds, the `self.iter` would need a length of at least `usize::MAX/2`, which seems a bit outrageous.
Regardless, I've changed it to use saturating arithmetic, because we can't represent a character position > `usize::MAX`, and I'm not sure how much value it adds to rework things to support it.

apoelstra: If you use e.g. [`iter::Repeat`](https://doc.rust-lang.org/std/iter/struct.Repeat.html) the iterator will effectively have an infinite length and will always quickly return `Some` for `nth`.

Maybe we should write some tests that depend on this.

apoelstra: Ah, we need `ExactSizeIterator`. We can get this on `repeat([a, b]).take()` but annoyingly only since Rust 1.82. https://doc.rust-lang.org/std/iter/struct.Take.html#impl-ExactSizeIterator-for-Take%3CRepeat%3CT%3E%3E

So maybe we'll have to hold off on using this obscenity as a test construction.

Anyway, saturating arithmetic seems fine to me. I would've used checked arithmetic personally.

apoelstra (approved):
ACK 1ec196f257a64cf06e684d710d550fe9b7205fc1; successfully ran local tests

tcharding (approved):
ACK 1ec196f257a64cf06e684d710d550fe9b7205fc1

PR #236, opened 2026-04-17
Introduce `Char` enum and rewrite `BytesToHexIter` to return it

### LINKED ISSUE #234: Should `BytesToHexIter` be byte-level reverse, or char-level reverse?
At the moment, the `DoubleEndedIterator` impl on the `BytesToHexIter` is byte-level reversed. When going backwards, the hex chars are yielded in forward nibble order, but reverse byte order. See the following snippet from the `encode_iter_backwards` test case:
```rust
let bytes = [0xde, 0xad, 0xbe, 0xef];
let lower_want = "efbeadde";
let upper_want = "EFBEADDE";
```

Is this correct? Or, should the backwards pass through the iterator be an complete reverse of the forward pass (i.e. reversed byte and nibble order or "FEEBDAED" for the above)?
Regardless, I think this should be documented, because it confused me when I noticed it and it's not mentioned in the rendered docs.

mpbagot:
Currently, BytesToHexIter takes in u8 and returns char. By introducting an enum Char type, we can use that in place of char to improve type safety. Further, by returning pairs of Chars, we can eliminate bugs associated with the internal buffer and allow users to trivially control character/nibble order during encoding.

- Patch 1 introduces the Char enum and replaces u8 in Table with it
- Patch 2 rewrites BytesToHexIter to return [Char; 2], including updating the benchmarks to suit.
- Patch 3 updates the api files.

Closes #234

### DIFF
--- a/api/all-features.txt
+++ b/api/all-features.txt
@@ -488,6 +488,77 @@ impl<T> core::clone::CloneToUninit for hex_conservative::Case where T: core::clo
 pub unsafe fn hex_conservative::Case::clone_to_uninit(&self, dest: *mut u8)
 impl<T> core::convert::From<T> for hex_conservative::Case
 pub fn hex_conservative::Case::from(t: T) -> T
+#[repr(u8)] pub enum hex_conservative::Char
+pub hex_conservative::Char::Eight = 56
+pub hex_conservative::Char::Five = 53
+pub hex_conservative::Char::Four = 52
+pub hex_conservative::Char::LowerA = 97
+pub hex_conservative::Char::LowerB = 98
+pub hex_conservative::Char::LowerC = 99
+pub hex_conservative::Char::LowerD = 100
+pub hex_conservative::Char::LowerE = 101
+pub hex_conservative::Char::LowerF = 102
+pub hex_conservative::Char::Nine = 57
+pub hex_conservative::Char::One = 49
+pub hex_conservative::Char::Seven = 55
+pub hex_conservative::Char::Six = 54
+pub hex_conservative::Char::Three = 51
+pub hex_conservative::Char::Two = 50
+pub hex_conservative::Char::UpperA = 65
+pub hex_conservative::Char::UpperB = 66
+pub hex_conservative::Char::UpperC = 67
+pub hex_conservative::Char::UpperD = 68
+pub hex_conservative::Char::UpperE = 69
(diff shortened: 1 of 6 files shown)


mpbagot:
It should be noted that I didn't introduce any of these traits for `Char` that Kix suggested (only `From` for `char` and `u8`):
```rust
// These are trivially correct `unsafe` casts
impl AsRef<str> for [Char] { /* ... */ }
impl<const N: usize> AsRef<str> for [Char; N] { /* ... */ }
impl AsRef<[u8]> for [Char] { /* ... */ }
impl<const N: usize> AsRef<[u8]> for [Char; N] { /* ... */ }
impl<const N: usize> AsRef<[u8; N]> for [Char; N] { /* ... */ }
// Borrows and mutable casts just as AsRef
impl<const N: usize> From<[Char; N]> for [u8; N] { /* ... */ }
```
These seem to have problems with orphan rules, since `[Char; N]` is considered a foreign type. Happy to look again if I've missed something obvious or if someone would like to do a follow-up for them.

apoelstra:
Frustrating. I think you're correct about the orphan rules.

apoelstra (approved):
ACK aef87f48076895e31f19eb0ed0f10cb085616dae; successfully ran local tests

tcharding (approved):
ACK aef87f48076895e31f19eb0ed0f10cb085616dae

tcharding:
> since `[Char; N]` is considered a foreign type

We could add a `Pair` type which would solve this problem and eliminate having to think which array index is the high nibble and which is the low?

apoelstra:
> We could add a `Pair` type which would solve this problem and eliminate having to think which array index is the high nibble and which is the low?

IMO this is too much abstraction. It will be harder for users to fight their way out of our types to get real data than it would be for them to just have to manually pull apart arrays.

I also think it's "obvious" that the first nybble is the high one and the second one is the low one. I don't think I've ever seen a byte vector hex-encoded any other way.

apoelstra:
aef87f48076895e31f19eb0ed0f10cb085616dae needs rebase

tcharding:
Works for me!

tcharding (approved):
ACK becc2de33d16b92edca095be6af97ddda06bc782

apoelstra (approved):
ACK becc2de33d16b92edca095be6af97ddda06bc782; successfully ran local tests

PR #237, opened 2026-04-19
Add rust-bitcoin.org as homepage

tcharding:
Add a homepage that will link to the soon-to-be-update website that documents this crate's existence.

ref: https://github.com/rust-bitcoin/rust-bitcoin.github.io/pull/45


apoelstra:
So, other than in the org name and a single doccomment in display.rs, the word "bitcoin" does not appear anywhere in this crate. We had a long discussion about the name and landed on `hex-conservative` rather than `bitcoin-hex` because the intention is that this be a general-purpose hex crate that just tries to have a small API and conservative MSRV.

So do we really want its official homepage to be the rust-bitcoin org page?

tcharding:
hmm, good points. I was only thinking from the other side, from rust-bitcoin.org it looked right. Thanks, closing.

PR #240, opened 2026-04-20
Revert "Make iter module public"

tcharding:
This reverts commit 508962bd30b70c5cce65971b1e69d4de2c4bba06.

We are now releasing `v1.1.0` from master branch so this branch will only be for security patches to `v1.0`. As such we don't want any other changes on the branch.

Over on master we kept the `iter` module private and re-exported the iterator types from the crate root.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -29,7 +29,8 @@ impl<'a> HexToBytesIter<HexDigitsIter<'a>> {
     ///
     /// If the input string is of odd length.
     #[inline]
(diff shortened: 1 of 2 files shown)


tcharding:
I already ticked of the list item in #209, if this gets shot down be sure to see to that item.

apoelstra:
Can you be more specific what this branch is for? Why don't we just delete it?

tcharding:
Ah yes I agree lets do that.

tcharding:
Done, github auto-closed this when I deleted the branch.

PR #242, opened 2026-04-21
Char improvements

Kixunil:
These are a few follow ups to #236 

I'm too tired for tests rn, will add them later.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,7 +225,7 @@ impl Case {
 //
 // The `repr(u8)` guarantees that representation matches the ASCII byte value of the character,
 // making transmute between `Char` and `u8` sound whenever the byte is a valid hex digit.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[repr(u8)]
 pub enum Char {
     /// `'0'`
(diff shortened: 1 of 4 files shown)


apoelstra:
utACK a3b664063da05d130ec2e4ecb85ed991a25212a0.

Arguably the "Because" lines in the doccomments are unnecessary. From a user point of view, these conversions make sense because we've provided a safe method that provides them. The user doesn't care what representation magic is needed to get it to work.

Tests would be nice, I guess. The code appears to be correct, though it's much more aggressively optimized than anybody else would do.

CI is failing because of a clippy lint (looks like one of your pointer casts has a convenience function now) and also the API files, which if they annoy you too much, me or Tobin can add.

Kixunil:
Addressed all the feedback, including adding tests.

apoelstra:
Can you run the formatter?

src/lib.rs
```diff
@@ -273,6 +274,52 @@ pub enum Char {
...
+    /// Casts a slice of `Char`s to `&[u8]`.
+    ///
+    /// This conversion is zero-cost.
+    #[inline]
+    pub fn slice_as_bytes(slice: &[Self]) -> &[u8] {
+        let ptr = slice.as_ptr().cast();
+        let len = slice.len();
+        // SOUNDNESS: `Self` is repr(u8)
+        // Because all chars are ASCII a slice of chars is also guaranteed to be valid slice of
+        // bytes.
```
tcharding, line 297: nit: I rekon its better without this comment. Mainly because 'all chars are ASCII' is not true if one interprets 'chars' as the Rust type. But also because the SOUNDNESS line already documents  the code.

tcharding (approved):
ACK 2713e69e321f9e6df32ccad9f724c53ebc51b319

tcharding:
Nice PR.

apoelstra (approved):
ACK 49818c6c6465f6b9b25b07838fac07ffe269b987; successfully ran local tests

tcharding (approved):
ACK 49818c6c6465f6b9b25b07838fac07ffe269b987

src/lib.rs
Kixunil: Maybe it should be `Char`s. Feel free to make a PR.

PR #243, opened 2026-04-29
Use `Char` table lookup to improve decode performance

mpbagot:
The decode performance is currently hampered by the hex_chars_to_byte function which converts u8 pairs to chars, then to u32s before combining them to a single u8 byte. Using the Char type, we can make a lookup table on the stack and dramatically improve decode performance.

- Patch 1 introduces a benchmark for decode_to_array and decode_to_vec.
- Patch 2 adds decode_nibble on Char to decode a u8 hex char to a nibble value and uses this for decoding in hex_chars_to_byte.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -275,6 +275,36 @@ pub enum Char {
 }
 
 impl Char {
+    /// Returns the nibble value (0–15) of this hex character.
+    #[inline]
+    pub(crate) fn decode_nibble(b: u8) -> Option<u8> {
+        // Each valid hex byte maps to its nibble value; 0xFF marks invalid entries.
+        // Char variant discriminants equal their ASCII byte values, so they index directly.
+        #[rustfmt::skip]
+        static TABLE: [u8; 256] = {
+            let mut t = [0xFF_u8; 256];
+            // Each Char variant is a u8. So all `as usize` casts are safe.
+            t[Char::Zero  as usize] = 0;  t[Char::One   as usize] = 1;
+            t[Char::Two   as usize] = 2;  t[Char::Three as usize] = 3;
+            t[Char::Four  as usize] = 4;  t[Char::Five  as usize] = 5;
+            t[Char::Six   as usize] = 6;  t[Char::Seven as usize] = 7;
+            t[Char::Eight as usize] = 8;  t[Char::Nine  as usize] = 9;
+            t[Char::LowerA as usize] = 10; t[Char::UpperA as usize] = 10;
+            t[Char::LowerB as usize] = 11; t[Char::UpperB as usize] = 11;
+            t[Char::LowerC as usize] = 12; t[Char::UpperC as usize] = 12;
+            t[Char::LowerD as usize] = 13; t[Char::UpperD as usize] = 13;
+            t[Char::LowerE as usize] = 14; t[Char::UpperE as usize] = 14;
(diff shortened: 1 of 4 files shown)


mpbagot:
The benchmark added in the first patch has the following change in performance from before to after the second patch:
```
decode_to_array/bytes/4 time:   [5.2166 ns 5.2205 ns 5.2255 ns]
                        change: [−12.583% −12.410% −12.268%] (p = 0.00 < 0.05)
                        Performance has improved.

decode_to_array/bytes/32
                        time:   [19.153 ns 19.166 ns 19.180 ns]
                        change: [−41.990% −41.935% −41.883%] (p = 0.00 < 0.05)
                        Performance has improved.

decode_to_array/bytes/256
                        time:   [149.56 ns 149.76 ns 149.94 ns]
                        change: [−43.702% −43.631% −43.561%] (p = 0.00 < 0.05)
                        Performance has improved.

decode_to_vec/bytes/4   time:   [11.492 ns 11.499 ns 11.504 ns]
                        change: [−8.3526% −8.2020% −8.0534%] (p = 0.00 < 0.05)
                        Performance has improved.

decode_to_vec/bytes/32  time:   [22.992 ns 23.014 ns 23.042 ns]
                        change: [−39.250% −39.201% −39.150%] (p = 0.00 < 0.05)
                        Performance has improved.

decode_to_vec/bytes/256 time:   [146.21 ns 146.38 ns 146.61 ns]
                        change: [−45.562% −45.470% −45.366%] (p = 0.00 < 0.05)
                        Performance has improved.
```

src/lib.rs
```diff
@@ -275,6 +275,35 @@ pub enum Char {
...
+            t[Char::Eight as usize] = 8;  t[Char::Nine  as usize] = 9;
+            t[Char::LowerA as usize] = 10; t[Char::UpperA as usize] = 10;
+            t[Char::LowerB as usize] = 11; t[Char::UpperB as usize] = 11;
+            t[Char::LowerC as usize] = 12; t[Char::UpperC as usize] = 12;
+            t[Char::LowerD as usize] = 13; t[Char::UpperD as usize] = 13;
+            t[Char::LowerE as usize] = 14; t[Char::UpperE as usize] = 14;
+            t[Char::LowerF as usize] = 15; t[Char::UpperF as usize] = 15;
+            t
+        };
+        let n = TABLE[b as usize];
```
tcharding:
```suggestion
        let n = TABLE[usize::from(b)];
```

tcharding: nit: Saves reader from having to wonder (or look up at the argument type) if any bits are cast away.

tcharding (approved):
ACK 2df48816493fa2894b324582658c4d5b3bfe0f11

tcharding:
Nice PR, like a real systems programmer.

tcharding (approved):
ACK 158a53821fe609e0f4e7f4bd4113d2ed4f5e0ef4

apoelstra (approved):
ACK 158a53821fe609e0f4e7f4bd4113d2ed4f5e0ef4; successfully ran local tests

Kixunil (commented):
One thing that's particularly confusing about this is the method being on `Char` while not having anything to do with `Char`. Maybe it would make sense to build a table of `Option<Char>` instead and then only cast it after calling the function? At least it removes one branch.

PR #245, opened 2026-05-09
macros: add docs and examples to fmt_hex_exact

alexgrad42:
Improve the macro by:
- Adding comprehensive doc-comments with a usage example.
- Explicitly documenting the return type and the static assertion.

Fixes: #244

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -254,7 +254,11 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { self.display(f, Case::Upper) }
 }
 
-/// Format known-length array as hex.
+/// Efficiently formats a sequence of bytes as a hexadecimal string with compile-time-known length.
+///
+/// Whenever the length of a sequence of bytes (usually an array, but it could be any iterator)
+/// is known at compile time, it is more efficient to use this macro because various length checks
+/// and loops are elided. The buffer is just filled once and then emitted to the formatter.
 ///
 /// This supports all formatting options of formatter and may be faster than calling `as_hex()` on
 /// an arbitrary `&[u8]`. Note that the implementation intentionally keeps leading zeros even when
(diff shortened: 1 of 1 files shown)


src/display.rs
```diff
@@ -270,14 +273,46 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
...
+///         // Explicitly specifying Upper case (4 arguments)
+///         fmt_hex_exact!(f, 32, &self.0, Case::Upper)
+///     }
+/// }
+/// ```
 #[macro_export]
 macro_rules! fmt_hex_exact {
+    ($formatter:expr, $len:expr, $bytes:expr) => {
+        $crate::fmt_hex_exact!($formatter, $len, $bytes, $crate::Case::Lower)
+    };
```
Kixunil: In #244 we agreed this is not worth it, so please remove it.

src/display.rs
```diff
@@ -254,7 +254,10 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { self.display(f, Case::Upper) }
 }
 
-/// Format known-length array as hex.
+/// Formats a fixed-length byte array as a hexadecimal string.
+///
+/// This macro provides a convenient way to format data as hex while ensuring
+/// at compile-time that the provided data matches the expected length.
```
Kixunil: I think this should be something like:

Efficiently formats a sequence of bytes with compile-time-known length.

Whenever the length of a sequence of bytes (usually an array, but it could be any iterator) is known at compile time, it is more efficient to use this macro because various length checks and loops are elided. The buffer is just filled once and then emitted to the formatter.

src/display.rs
```diff
@@ -270,14 +273,46 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
...
+///   Defaults to `Lower` if omitted.
+///
+/// ## Returns
 ///
-/// ## Panics
+/// Returns `core::fmt::Result`.
+///
+/// # Static Assertions
 ///
 /// This macro panics if `$len` is not equal to `$bytes.len()`. It also fails to compile if `$len`
```
Kixunil: This is confusing. The panic when lengths are unequal is not a static assertion but actual runtime panic. Only the max length is a static assertion.

Kixunil (changes requested):

src/display.rs
apoelstra: I'd rather just add noise to all my own code than argue about this, but FWIW we didn't agree that this isn't worth it. Only that the case argument has positive value.

src/display.rs
apoelstra: Yeah, I think we should add a whole section on static assertions (and explicitly say that the static assertion is why we have a macro rather than a function).

src/display.rs
alexgrad42: Done

src/display.rs
alexgrad42: Done

src/display.rs
alexgrad42: Done

src/display.rs
```diff
@@ -272,10 +276,49 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
...
+/// use std::fmt;
+///
+/// struct MyHash([u8; 32]);
+///
+/// impl fmt::Display for MyHash {
+///     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+///          // Defaulting to Lower case
+///          fmt_hex_exact!(f, 32, &self.0, Case::Lower)
+///      }
+///  }
```
Kixunil: I'm worried that this impl will lead people to repeat this exact code, which might be non-ideal. I believe it's better to just forward `Display` to `LowerHex` by calling `fmt::LowerHex::fmt`. So I suggest either remove `Display` or call into `LowerHex` to show the better way.

Kixunil (approved):
There's still room for improvement but I don't find it super important, so

ACK b235ec3142a5dbb01bde2e5e8f05e21149a7327a

But if you have the time and desire to apply the suggestion I'll be happy to re-review.

src/display.rs
tcharding: Yeah I'd also prefer if the `Display` iml was removed. It doesn't add any additional information.

src/display.rs
```diff
@@ -272,10 +276,49 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
...
 ///
-/// This macro panics if `$len` is not equal to `$bytes.len()`. It also fails to compile if `$len`
-/// is more than half of `usize::MAX`.
+/// This macro panics if `$len` is not equal to `$bytes.len()`
+///
+/// ## Static Assertions
+///
+/// The use of a macro instead of a function allows for compile-time length validation.
+/// This macro fails to compile if `$len` is more than half of `usize::MAX`. This prevents runtime panics or logic errors when formatting fixed-size
+/// primitives like Bitcoin hashes or public keys.
```
tcharding: Can you put all these lines together and wrap at 100 characters please.

alexgrad42: Done

src/display.rs
alexgrad42: Done

apoelstra (approved):
ACK cec20392ac7ddf7b88f41f2f0591a8b51c257c5c; successfully ran local tests

Kixunil (approved):
ACK cec20392ac7ddf7b88f41f2f0591a8b51c257c5c

PR #247, opened 2026-05-11
Add fmt_hex_exact_lower and fmt_hex_exact_upper macros

alexgrad42:
The implementation wraps the existing `fmt_hex_exact!` macro to minimize code duplication and keep it clean.

Fixes: #246

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -282,7 +282,7 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
 ///
 /// ## Panics
 ///
-/// This macro panics if `$len` is not equal to `$bytes.len()`
+/// This macro panics if the length of the encoded item is larger than `$len`.
 ///
 /// ## Static Assertions
 ///
@@ -293,7 +293,7 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
 /// ## Examples
 ///
 /// ```rust
-/// use hex_conservative::{fmt_hex_exact, Case};
+/// use hex_conservative::{fmt_hex_max, Case};
 /// use std::fmt;
 ///
 /// struct MyHash([u8; 32]);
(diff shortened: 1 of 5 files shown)


src/display.rs
```diff
@@ -280,10 +280,6 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
 ///
 /// Returns [`core::fmt::Result`].
 ///
-/// ## Panics
-///
-/// This macro panics if `$len` is not equal to `$bytes.len()`
-///
```
Kixunil, line 286: This shouldn't be removed, rather it should be updated to say that it will panic if the length of the encoded item is larger than `$len`.

src/display.rs
```diff
@@ -293,37 +289,74 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
...
 /// }
 /// ```
 #[macro_export]
-macro_rules! fmt_hex_exact {
+macro_rules! fmt_hex_max {
     ($formatter:expr, $len:expr, $bytes:expr, $case:expr) => {{
         // statically check $len
         #[allow(deprecated)]
         const _: () = [()][($len > usize::MAX / 2) as usize];
+        $crate::display::fmt_hex_max_fn::<_, { $len * 2 }>($formatter, $bytes, $case)
```
Kixunil, line 327: Length check is missing here.

Kixunil (changes requested):

src/display.rs
alexgrad42: Done,  please check

src/display.rs
alexgrad42: Done, please check

Kixunil (approved):
ACK a9d20327414a8c90745b144574067ee6ddc02190

src/display.rs
```diff
@@ -293,37 +293,80 @@ impl fmt::UpperHex for DisplayByteSlice<'_> {
...
+/// Formats bytes as hex in lower case.
+///
+/// See [`fmt_hex_max!`] for details.
+#[macro_export]
+macro_rules! fmt_hex_lower {
+    ($formatter:expr, $len:expr, $bytes:expr) => {
+        $crate::fmt_hex_max!($formatter, $len, $bytes, $crate::Case::Lower)
+    };
+}
+pub use fmt_hex_lower;
```
tcharding, line 357: Do we definitely want this re-export? I see we have it for `fmt_hex_exact` and its released so we cannot remove it. But it looks wrong/unnecessary. The macro is public so already available at the crate root.

tcharding (approved):
ACK a9d20327414a8c90745b144574067ee6ddc02190

src/display.rs
apoelstra: I think it's fine to keep it for consistency with `fmt_hex_exact`, and also because I really don't care to look up what the macro exporting rules are in this edition of Rust.

apoelstra (approved):
ACK a9d20327414a8c90745b144574067ee6ddc02190; successfully ran local tests

src/display.rs
tcharding: No sweat, I checked and all the macros in `display` are uniform in this respect. Onwards and upwards.

PR #257, opened 2026-08-06
Patch sub-errors to make them backwards compatible

tcharding:
We would like to alias the `DecodeFixedLengthBytesError` back into `0.2` and `0.3` as `HexToArrayError` [0].

During the 1.0 dev cycle we made a few minor API breaks. Turns out we were probably overly savage and the benefits of re-instating the previous API outweigh the maintenance risks we were attempting to mitigate.

Put the `InvalidLengthError` and `InvalidCharError` back to how they were when in the `0.3` release.

Done as part of #254

[0] See issue #254 for full context and justification.


apoelstra (approved):
ACK 828b93d744b814eed633b6bfc2734ed7058e8009; successfully ran local tests

apoelstra:
@tcharding are you willing to make a release PR off `master`? Then we can double-back and do a new 0.2.x release with the semver trick. This should all be straightforward now that CI is passing there.

tcharding:
Yep can do.

